PR #6, opened 2020-03-17
Add columns 'hospitalized' and 'recovered'

jb3-2:
@tlorusso I have added two more columns for which BS published data today. Do you agree? If yes, please merge - otherwise let's discuss!


rsalzer:
Then please also merge the Zurich data to the cantonal file ;-)

jb3-2:
@rsalzer Currently @tlorusso is optimizing the machinery. If I understand him correctly, we will have one file per canton, and all of those will be merged into the CH file. This way we will have less manual work (update data only in the canton file), and less risk of inconsistencies. Would that help you?

Issue #22, opened 2020-03-20
einheitliche Datenfiles der Kantone


zdavatz:
Besteht die Möglichkeit, dass wir für alle Kantone ein einheitliches CSV-Format kriegen? Am besten wäre ein Tool, welches das Format validiert vor dem Upload.

tlorusso:
Ja, hat hohe Prio! Sollten das Format heute angleichen (inkl.
Beispiel-File). Hast du bezüglich Validierung einen Vorschlag? WIr wollen
die Hürden so tief wie möglich behalten, 'Validierung' ginge auch bei
Sichtung der Pull-Requests.


*Thomas Lo Russo*
*th.lorusso@gmail.com <th.lorusso@gmail.com>*


Am Fr., 20. März 2020 um 08:57 Uhr schrieb Zeno R.R. Davatz <
notifications@github.com>:

herrstucki:
Es wäre wahrscheinlich nicht zu aufwändig, das mit einer GitHub Action und z.B. [CSVLint](https://csvlint.io/about) für PRs zu automatisieren.

zdavatz:
Ich würde irgendein Bash-Script machen. Seit ihr auf Linux?

zdavatz:
Könnt ihr bitte diese Vorlagen verwenden zum abfüllen der Daten pro Kanton?

https://github.com/zdavatz/covid19_ch/blob/master/data-cantons-csv/dd-covid19-ch-cantons-latest.csv

hier noch das File für die ganze Schweiz

https://github.com/zdavatz/covid19_ch/blob/master/data-switzerland-csv/dd-covid19-ch-switzerland-latest.csv

zdavatz:
Hier noch das README dazu: https://github.com/zdavatz/covid19_ch/blob/master/README.md#data-per-canton

zdavatz:
@herrstucki warum wäre das zu aufwendig? Finde ich eine sehr gut Idee!

zdavatz:
https://stackoverflow.com/questions/33523362/how-to-compare-two-csv-files-in-windows/45591193

zdavatz:
Hier ein `awk` Befehl für Linux, welcher die Headers von zwei CSV Files vergleicht und die Headers anzeigt wenn sie übereinstimmen:
[...]

PR #24, opened 2020-03-20
Csv validation

rokroskar:
* adds a github action to validate CSV files
* fixes missing fields in TI data

Since this is a fork, the action won't run on the PR - but you can see it here: https://github.com/rokroskar/covid_19/runs/521628375?check_suite_focus=true

addresses #22


rokroskar:
Actually - if this is to be a required status check to pass we probably want to run it on every push, not just on changes to csv files... I will make the change

fabian:
Sample output: https://github.com/rokroskar/covid_19/runs/521628375?check_suite_focus=true#step:3:11

PR #38, opened 2020-03-20
updating VS with hist

zukunft:


tlorusso:
Great! 😁

zukunft:
We need to show the what Swiss quality is. At the moment there is some upward potential until we reach the level of e.g. Korea. Thanks for the initiative !

PR #39, opened 2020-03-21
Update for SZ

ebeusch:
used cantonal press releases as data source (see file)
fyi: it looks to me like ncumul_tested is not defined clearly enough. (i.e. ncumul within day or over time?)


tlorusso:
So far, its cumulative over time, as this is the way the few Cantons that publish this number are reporting it. Hopefully, further will follow. However, there might be some differences in how they report (daily vs cumulative), so you're right : we should provide a very precise definition to assure the figures will be collected as coherently as possible.

ebeusch:
OK, that's what I thought but I've noticed inconsistencies in some files. We should check and clean for those thus.

zdavatz:
How many cantons are still missing? Could we eliminate one?

PR #47, opened 2020-03-21
Update COVID19_Fallzahlen_Kanton_LU_total.csv

rsalzer:
change dates to ISO


rsalzer:
there is already another pull request with the same changes... closing

Issue #40, opened 2020-03-21
Rmd or Jupyter notebooks


betatim:
Does someone know of Rmd or Jupyter notebooks that read and visualise the data?

If yes it could be nice to include one or two of them here. We can also add a mybinder.org link to let people run them.

Great work!

jb3-2:
Maybe one of those here is close to what you're looking for? https://db.schoolofdata.ch/project/65

rokroskar:
@betatim we started a project on the Renku platform for this: https://renkulab.io/projects/covid-19/covid-19-public-data

The notebook for this data is here: https://renkulab.io/projects/covid-19/covid-19-public-data/files/blob/notebooks/examples/openzh-covid-19-example.ipynb

We would be very happy if people contributed more visualizations and analysis! In particular, I would love to scale these by population data...

baryluk:
I just tested the Jupyter notebook, and it works really nice, and is easy to use. Good job.

zdavatz:
@rokroskar Where do you get the population data from?

rokroskar:
@zdavatz good question, it took me a while to find the right table. It's from the [Bundesamt für Statistik](https://www.bfs.admin.ch) and I got it from a data request form that now I cannot for the life of me locate again. This is the link it was obtained from: https://www.bfs.admin.ch/bfsstatic/dam/assets/9635943/master

However, I just now found a better document: https://www.bfs.admin.ch/bfs/de/home/statistiken/bevoelkerung/stand-entwicklung/raeumliche-verteilung.assetdetail.9486060.html

All the available data that I could find is from 2018 unfortunately... if you have a better source please let me know!

rokroskar:
Ah I just realized those don't have the same data - the first one is reported on 31.12.2018 the second one on 1.1.2018. 🤷‍♂

metaodi:
See #246 for a Rmd document

ivanek:
The Rmd is now at #246

rokroskar:
[...]

PR #79, opened 2020-03-21
Add visualisation notebook

betatim:
This PR adds a notebook that visualises some of the data and computes how many days it took for the number of deaths/confirmed cases to double.

Also added a Binder badge to make it easy to run the notebook: https://mybinder.org/v2/gh/betatim/covid_19/visualisation?filepath=visualise.ipynb

ref: #40


metaodi:
Thanks a lot @betatim

PR #90, opened 2020-03-22
Update COVID19_Fallzahlen_Kanton_TI_total.csv

hroest:
correct number of dead, use stable URL


hroest:
also compare to: https://de.wikipedia.org/wiki/COVID-19-Pandemie_in_der_Schweiz#Todesf%C3%A4lle

PR #95, opened 2020-03-22
adding hist GE

zukunft:


zukunft:
Are the columns in the GE file correct? I have filled the missing history based on the existing data, but it looks to me like ncumul_vent, ncumul_ICF and ncumul_ICU_intub are mixed up. Can someone check, please?

Issue #61, opened 2020-03-21
A quick script for computing latest totals


baryluk:
In the directory with data:

```shell
for f in *.csv; do awk -F , '{if ($5) { print $1, $3, $5; }}' "$f" | tail -1; done | awk 'BEGIN { sum = 0; } { sum += $3; } END { print sum; }'
```
6262

Getting latest confirmed cases by subdivision:

```shell
for f in *.csv; do awk -F , '{if ($5) { print $1, $3, $5; }}' "$f" | tail -1; done |  sort -r -n -k 3
```

```
2020-03-20 VD 1432
2020-03-21 TI 918
2020-03-20 GE 873
2020-03-20 ZH 773
2020-03-20 BE 377
2020-03-21 BS 299
2020-03-21 BL 282
2020-03-20 VS 282
2020-03-20 GR 213
2020-03-20 AG 165
2020-03-20 NE 159
2020-03-21 LU 109
2020-03-20 SG 98
2020-03-21 TG 56
2020-03-20 ZG 48
2020-03-20 FL 40
2020-03-20 JU 29
2020-03-20 NW 28
2020-03-19 GL 17
2020-03-20 SH 14
2020-03-15 SZ 13
2020-03-21 UR 12
2020-03-18 AR 11
2020-03-09 FR 11
2020-03-14 AI 2
2020-03-13 OW 1
```

Feel free to include it in the repo, or readme. Public Domain. Signed off, Witold Baryluk.

tlorusso:
Might indeed make sense to include this in the readme, especially if we could 'refresh' it automatically after a file in the 'total'-folder is updated.

baryluk:
If you want I can make the second script output in more Markdown-friendly format, so it can be easily copy pasted into README.md on updates or something.

zdavatz:
nice work!

baryluk:
`../scripts/latest_per_canton.sh` output:

| Canton | Confirmed cases | Last update |
|:------:| ---------------:|:----------- |
|   VD   |            1432 | 2020-03-20  |
|   TI   |             918 | 2020-03-21  |
|   GE   |             873 | 2020-03-20  |
|   ZH   |             773 | 2020-03-20  |
|   BE   |             377 | 2020-03-20  |
|   BS   |             299 | 2020-03-21  |
|   VS   |             282 | 2020-03-20  |
|   BL   |             282 | 2020-03-21  |
|   GR   |             213 | 2020-03-20  |
[...]

PR #109, opened 2020-03-23
Add simple scripts to compute totals and per-canton latest commulative confirmed cases

baryluk:
Closes: https://github.com/openZH/covid_19/issues/61


zdavatz:
great work!

PR #121, opened 2020-03-23
Fix scraper for GE

baryluk:
Fix date format, and the fact that text in PDF sometimes is wrapped into
multiple lines.
    
Also numbers above 999 are printed with decimal separator, like 1'231, so
get rid of apostrophes before using sed.


metaodi:
Looks good.

Btw: I sucessfully integrated one of your scrapers with GitHub Actions, so that it updates the repo. Will now start doing this for all scrapers that run more or less stable :wink:

PR #122, opened 2020-03-23
Don't call curl directly, use a wrapper.

baryluk:
This will use curl, or wget, or GET automatically and set needed flags.

It will also open possibility of automatically archiving all downloaded
data for reference and running tests, fixing broken scraper,
calling web.archive.org to do a snapshot on their side, etc.


metaodi:
`./download.sh` only works if you call the script from it's directory, i.e. calling the script from the root directory of the repo with `scrapers/scrape_ar.sh` will fail. 

In `run_scraper.sh` I defined $DIR:
https://github.com/openZH/covid_19/blob/60fa566eb917fa7a6af457a46d4290aa85a2174b/scrapers/run_scraper.sh#L12

And then used $DIR instead of `./`. WDYT?

baryluk:
> `./download.sh` only works if you call the script from it's directory, i.e. calling the script from the root directory of the repo with `scrapers/scrape_ar.sh` will fail.
> 
> In `run_scraper.sh` I defined $DIR:
> 
> https://github.com/openZH/covid_19/blob/60fa566eb917fa7a6af457a46d4290aa85a2174b/scrapers/run_scraper.sh#L12
> 
> And then used $DIR instead of `./`. WDYT?

Yes, that makes sense. I need to go sleep. I will fix it tomorrow.

I also have a modification for download.sh, that allows it to ask web.archive.org to do a snapshot of the source page. It works. :) Will be in a next PR.

baryluk:
Ok. I think this whole `DIR` business (a lot of redundancy) is a bit ugly, but it is done, and works.

metaodi:
Thank you! Maybe there is a better way, but I'm not proficient enough in shell programming to propose something else.
I'll try to get more scrapers up and running based on this wrapper.

metaodi:
Currently download.sh simply chooses curl, wget and GET based on being available in the path. How does this help with sites like GL or SZ that do not accept curl? Am I missing something?

baryluk:
> Currently download.sh simply chooses curl, wget and GET based on being available in the path. How does this help with sites like GL or SZ that do not accept curl? Am I missing something?

We are setting user-agent to a fake one, so they don't know we are using curl.

The primary purpose of download.sh is to move this logic out of individual scripts,
but also enable automation (archiving, testing, logging of failures).

metaodi:
Perfect, thanks for explaining!

PR #123, opened 2020-03-23
Replace COVID-19 with SARS-CoV-2

### LINKED ISSUE #96: Terminology clarification
Dear all,

thanks a lot for keeping this repo up-to-date! There seems to be some confusion regarding the terminology: the infection rates refer to the infection with SARS-Cov-2, whereas the disease caused by the virus is called Covid-19.

Best wishes,
-Filippo

filippo82:
Replace COVID-19 with SARS-CoV-2 when referring to the infection cases.

Fixes #96

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,11 +4,15 @@
 [![GitHub commit](https://img.shields.io/github/last-commit/openZH/covid_19)](https://github.com/openZH/covid_19/commits/master)
 [![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/openZH/covid_19/master?filepath=visualise.ipynb)
 
-# COVID-19 Cases communicated by Swiss Cantons and Principality of Liechtenstein (FL)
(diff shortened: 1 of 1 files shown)


metaodi:
I'm not sure I understand. People get infected by the virus called "SARS-CoV-2". But if they get tested and are a "confirmed" case, that means they have COVID-19 (the desease the virus causes). Or am I missing something?

filippo82:
Hi @metaodi, the test only confirms that a person is infected by SARS-CoV-2. One can be infected but not develop the disease. SARS-CoV-2 (the species of virus) gives you COVID-19 (the illness) - the World Health Organisation explains it as being the same way you would say that HIV gives you AIDS. This [link](https://gd.zh.ch/internet/gesundheitsdirektion/de/themen/coronavirus.html#title-content-internet-gesundheitsdirektion-de-themen-coronavirus-jcr-content-contentPar-textimage_7) also states: "Im Kanton Zürich sind zurzeit 1068 Personen positiv auf das Coronavirus getestet worden."

metaodi:
@filippo82 thanks for the explanation!

PR #124, opened 2020-03-23
Add more scrapers to GitHub

metaodi:
This PR contains all scrapers I consider "simple", i.e. that are running stable and updating the same values that are already there.
There are more scrapers for cantons with extended schema (more columns after `source`). This need some more handling to not lose data. Then there are scrapers that run perfectly, but only the source CSV contains more information than just the `confirmed` and `deceases` case numbers.


scrapers/scrape_vs.sh
```diff
@@ -14,4 +14,4 @@ echo -n "Confirmed cases: "
 echo "$d" | egrep "bestätigte" | sed -E -e 's/^.*es ([0-9]+) best.* Fälle .*$/\1/'
 
 echo -n "Deaths: "
-echo "$d" | egrep "Tod" | sed -E -e 's/^.*Tod von ([0-9])+( |&nbsp;)Person.*$/\1/i'
+echo "$d" | egrep "Tod" | sed -E -e 's/^.*Tod von ([0-9]+)( |&nbsp;)Person.*$/\1/i'
```
baryluk: Good catch!

PR #142, opened 2020-03-24
Update COVID19_Fallzahlen_Kanton_SG_total.csv

hMutzner:
There are two values on 2020-03-04
the second case was announced 2020-03-06
see source
I don't know about time.


metaodi:
@hMutzner good catch!

PR #143, opened 2020-03-24
exit(1) when parsing the scraper output fails

metaodi:
To detect failures like [this](https://github.com/openZH/covid_19/runs/530757158?check_suite_focus=true), it's important that the parsing script fails with exit(1) when an error occurs.


baryluk:
Afaik, if the script fails and throws an exception it does call `exit(1)` automatically. I think this change is unnecessary.

Issue #152, opened 2020-03-24
Port scrapers from bash to Python


baryluk:
Bash is nice and it was quick to add many scraper, but we do see many failures, and it leads to some issues.

1. Not many people are familiar with bash, grep, sed, awk, to help with fix contribution.
2. Currently we use `set -e` (without `set -x`), which means if the scraper fails, it leaves no useful error messages.
3. Due to nature of `sed`, it often needs to be protected by extraneous and somehow redundant `grep` or `head -1` around it. This is because if the `sed` doesn't match anything on the line, it will output the line as is, malforming the output.

Use of Python (we can port one scraper at a time) will help:

1. More familiar.
2. Easier use of functions, logging, and exception messages, which should help with debugging.
3. Instead of using `grep+sed+tricks`, or `sed` with `s/^.*....().....*$/\1/` trickery, we can just use something like `re.search("...()...", d)[1]`, or a helper function (shared by scrapers) to make it ever easier. This should actually make shorter and cleaner script (no redundant `grep+sed`), that is also a bit more robust against parsing of multiple lines (no need for `head -1`, and eliminate some other steps that we do in bash right now.
4. Python has raw strings, which helps with regexp writing, due to reduction of backslash escaping needed.

zdavatz:
Ok! Sounds like a plan although `awk` is great, as Linus says ;).

metaodi:
The scraper for SG that is run by GitHub actions is already written in Python: https://github.com/openZH/covid_19/blob/master/scrapers/scraper_canton_sg.py

And i still have some floating around from morph.io, so I can simply move them here.

baryluk:
The script you have is not just a scraper (it does few other things), and is 5 times longer than needs to be. The same can be achieved in just few lines of code.

I will port most of he bash scrapers to Python today, it is a very easy process. Nothing fancy.

metaodi:
@baryluk I agree it's longer than needed (the scraper part being from L35-L52). But IMHO the sqlite part really makes sense to manage input and output:

- it defines a clear schema
- I can add constraints (like a UNIQUE) and therefore make sure to only generate one entry per date + time
- make sure the data in the CSV file is always used as an input (and if it's invalid it breaks)

Feel free to rewrite it shorter, but please be aware that we need more than "just a scraper", the result must be added to the existing file without losing information (e.g. if someone manually added data to a column the scraper does not provide).

baryluk:
[...]

PR #158, opened 2020-03-24
Port scrapers from bash to Python: library and ZH

baryluk:
This adds a port from bash to Python,
and a small library of functions to be used by scrapers.

So far ported just ZH, to see how it plays and if is reliable.
This also adds a 'Deaths' field.

Output is compatible with existing parser.

So far the code looks cleaner and kind of shorter than bash version,
which is nice.

This is a part of https://github.com/openZH/covid_19/issues/152

In the future, we probably should actually remove the `timestamp()`
and do it automatically in download.

I am leaving the file extension `.sh` for a moment, as this is expected by
some existing scripts probably, like `meta_scrape.sh`.
It is not consequential, beyond syntax highlighting in some editors.


baryluk:
I also implemented just scraper for AI, using this framework. It was a super short script. I will send a separate PR for it.

I like these functions. :)

scrapers/scrape_common.py
```diff
@@ -0,0 +1,28 @@
...
+
+import datetime
+import os
+import subprocess
+import re
+
+def download(url, encoding='utf-8'):
+  """curl like"""
+  downloader = os.path.join(os.path.dirname(__file__), 'download.sh')
+  return subprocess.run([downloader, url], capture_output=True).stdout.decode(encoding)
```
metaodi: Rather than spawning a subprocess, I'd suggest to use the requests library, it works cross-platform, provides nice error handling (like getting the HTTP Status code if something fails) etc. the download.sh was nice for the shell scripts, but is imho not the right tool for a python-based scraper.

scrapers/scrape_zh.sh
```diff
@@ -1,14 +1,19 @@
-#!/bin/sh
-set -e
+#!/usr/bin/env python3
 
-DIR="$(cd "$(dirname "$0")" && pwd)"  # " # To make editor happy
+from scrape_common import *
```
metaodi: Please don't `import *` it makes it unclear what's happening and pollutes the namespace (e.g. `filter` is part of the python standard library, so it's not possible to call right now). I'd suggest to use something like `import scrape_common as sc` and then call all the imported functions as `sc.download()` and `sc.filter()`

scrapers/scrape_common.py
baryluk: This was done intentionally to share the downloader.

scrapers/scrape_zh.sh
baryluk: `import *` is fine. It is under our control, so there is nothing wrong with it.

scrapers/scrape_common.py
metaodi: Fair point, we can refactor this later. 

From an error handling point-of-view it's so much more convenient to something like requests to download the content.

scrapers/scrape_zh.sh
metaodi: It's considered an anti-pattern that's super easy to avoid in this case. Please use `import scrape_common as sc`.

As said before: you are blocking the use of the standard library function `filter`, which is **not** obvious for others developers.

baryluk: It is not an anti-pattern. Otherwise this construct simply would not be in the language. It is only anti-pattern when you do `import *` from a package you don't have control off, and there is a risk of a new function/class overriding somebodies else class and breaking it in the future. There are marginal argument to be made about doing `import *` on many modules, which would make it harder to figure out which module each function is coming from, but again, that is not the case here.

I don't have time to argue about it tho, so I will change it, even if I strongly disagree with it.

baryluk:
PTAL

metaodi:
Thank you!

baryluk:
@metaodi Thank you.

Issue #162, opened 2020-03-25
Add historized data from official source JU


andreasamsler:
https://www.jura.ch/fr/Autorites/Coronavirus/Chiffres/Evolution-des-cas-COVID-19-dans-le-Jura.html --> voir details

zukunft:
See merge request #167

tlorusso:
Done. GRAZIE @zukunft

PR #167, opened 2020-03-25
update GE case hist base on report 24.03.2020

zukunft:


zukunft:
... and "Add historized data from official source JU" #162
... and update FL from PDF

zukunft:
Should I move "cleanup" the changes by using a branch?

PR #168, opened 2020-03-25
Update COVID19_Fallzahlen_Kanton_VS_total.csv

hMutzner:
delete one of the two records on 2020-03-24


metaodi:
Thanks, this was an error from our scraper. It's a ready fixed!

PR #172, opened 2020-03-25
Update Canton GE as of 25/03/2020

qligier:
Update Canton GE status from the usual source (https://www.ge.ch/document/covid-19-situation-epidemiologique-geneve/telecharger). I've not added the _ncumul_ICU_intub_ count because it may be false in the official report (going down from 41 to 2 in one day is a bit suspect).


tlorusso:
Sorry - we didn't check vor PRs this time before updating by ourselfes. Otherwise we would've merged. Closing this.

PR #174, opened 2020-03-25
Make sure there is only one entry per day

metaodi:
This only applies to CSVs that are updated via scraper. While it's
technically possible to have multiple values per day, it's hard to
ensure correctly from the scrapers. Some scrapers may not be able to
determine the time and will default to an empty string. If there is
already an entry with a time in place, this could lead to duplicate
entries.

### DIFF
--- a/scrapers/populate_database.py
+++ b/scrapers/populate_database.py
@@ -37,7 +37,7 @@ try:
             ncumul_released integer,
             ncumul_deceased integer,
             source text,
-            UNIQUE(date, time, abbreviation_canton_and_fl)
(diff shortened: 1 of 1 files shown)


zukunft:
Agreed. I think, I had at least one case, where I had one day with two times. Maybe we can build the scrapers, so that it only needs a single entry for the day to add. This allows to keep hist entries with two times.

metaodi:
I think consensus is to keep only one entry, so if it later changes, just update the existing row. This is currently out-of-scope for the scrapers. They only add a line if there is no entry for the day.

Issue #157, opened 2020-03-24
OW scraper failed


baryluk:
```
OW
Scraped at: 2020-03-24T21:15:10+01:00
Date and time: 24.03.2020
Confirmed cases:
```

Change of wording, on website. now is:

```
Fallzahl Kanton Obwalden

Bisher ist bei 25 Personen im Kanton Obwalden das Coronavirus nachgewiesen worden (Stand: 24. März 2020).
```

Easy fix. On it.

baryluk:
Actually, this is a bit weird, I think it should been working existing scraper:

```bash
echo "$d" | egrep "ist bei [0-9]+ Personen" | sed -E -e 's/^.*ist bei ([0-9]+) Personen.*$/\1/'
```

I might be blind, or maybe there is some extra space / non-breaking space.

Checking.

baryluk:
Oh. This looks to be because it is using `windows-1252` encoding in HTML, but we expect UTF-8, so grep then is confused by `M�rz`.

baryluk:
Fix available. (just use `iconv`). PR coming soon.

PR #177, opened 2020-03-25
Convert encoding of OW page to UTF-8

baryluk:
OW site apparently is using non-UTF-8 encoding as of 2020-03-24.
Not sure how it was before.

So convert it manually to UTF-8. Otherwise, grep will be complaining and
not output proper output, as it will think the file is binary.

Once we port all to Python, we could be smarter, and detect encoding
dynamically based on HTML headers probably.

Closes: https://github.com/openZH/covid_19/issues/157

### DIFF
--- a/scrapers/scrape_ow.sh
+++ b/scrapers/scrape_ow.sh
@@ -4,7 +4,7 @@ set -e
 DIR="$(cd "$(dirname "$0")" && pwd)"  # " # To make editor happy
 
 echo OW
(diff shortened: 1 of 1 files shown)


metaodi:
I want to merge this. Please remove all those "data" commits.

You could disable the GitHub actions on your fork or create a separate branch for code changes to avoid this.

baryluk:
@metaodi Ooo. I didn't know actions are automatically enabled on fork. This explains why I constantly had merge and conflicts issues constantly.

I did disable actions in my fork now, and cleaned up commits, so it should now merge cleanly.

PTAL

Issue #169, opened 2020-03-25
Show URL in Scraper output


zdavatz:
@baryluk can you please also show the URL per canton your are scraping in the output of https://www.functor.xyz/covid_19/scrapers/outputs/latest.txt or if you switch to python, show it in the output of the python-scraper. That would be really useful. 

@all please be aware that the URLs of the websites from the cantons in your daily stats [files](https://github.com/openZH/covid_19/tree/master/fallzahlen_kanton_total_csv) with the source of the information shows up on the left side of the Dashboard: http://covid19.ddrobotec.com/

So if everything is working correctly and is in sync, then the numbers here: https://www.functor.xyz/covid_19/scrapers/outputs/latest.txt and here http://covid19.ddrobotec.com/ should be in sync, including the URLs on the left of our dashboard.

baryluk:
Can be done.

baryluk:
It can be done very easily in a Python one. A little tricky, but doable in the bash.

I will add it in few moments.

zdavatz:
Thank you!

baryluk:
@zdavatz Could you verify if what is being output from scrapers, meta_scrape and parser is ok for you?

zdavatz:
@baryluk yes, this looks good. Only thing that is missing is the timestamp if a scraper fails. Please always print the timestamp Failed or not Failed should not matter. Please do not output error messages.

```
~/.software/covid_19/scrapers> ./meta_scrape.sh 
AG 2020-03-25T15:15     319       2 OK 2020-03-26T09:28:32+01:00 https://www.ag.ch/de/themen_1/coronavirus_2/alle_ereignisse/alle_ereignisse_1.jsp
AI 2020-03-28T17:00       9       - OK 2020-03-26T09:28:33+01:00 https://www.ai.ch/themen/gesundheit-alter-und-soziales/gesundheitsfoerderung-und-praevention/uebertragbare-krankheiten/coronavirus
AR 2020-03-26T09:00      40       2 OK 2020-03-26T09:28:33+01:00 https://www.ar.ch/verwaltung/departement-gesundheit-und-soziales/amt-fuer-gesundheit/informationsseite-coronavirus/
BE 2020-03-26T          660       7 OK 2020-03-26T09:28:33+01:00 https://www.besondere-lage.sites.be.ch/besondere-lage_sites/de/index/corona/index.html
[...]

PR #190, opened 2020-03-26
Snapshot using web.archive.org on each scrape

baryluk:
Closes: https://github.com/openZH/covid_19/issues/111

PR #191, opened 2020-03-26
Show URL in scraper output

baryluk:
Only for Python scrapers for a moment.

Multiple Downloading could be present in the output of the scraper,
so capture them as a list in the parser.
    
Closes: https://github.com/openZH/covid_19/issues/169


baryluk:
Because I messed up branches, please only merge after https://github.com/openZH/covid_19/pull/190

Will try better next time to not include changes from other branches.

Thanks.

PR #184, opened 2020-03-25
Fixes SG scraper and add deaths readout (corrected)

deragent:
Fixes #182

Issue #182, opened 2020-03-25
Fix scraper fro SG (and add deaths)


deragent:
Format changed slightly.
They also now publish the number of deaths.

deragent:
I am working on this.

baryluk:
Good catch. Thanks.

PR #194, opened 2020-03-26
Pythonize all but one scrapers

baryluk:
The only remaining one is VD, which has another PR in progress
on it.

Pythonized:

AG
AR
BE
BL
BS
GE
GL
GR
JU
LU
NE
NW
OW
SG
SH
SO
TG
TI
UR
VS
ZG

There is no functional differences, maybe with exception of one or two
scrapers, which now also extract deaths.

Regexps are mostly ported irectly without changes, but in few places,
they were simplified, or made even more robust than before
(non-greedy matching, less specified separators, etc).

Added function `pdfdownload` to `scrape_common.py`, which download pdf using `downloader.sh` and converts to text using `pdftotext` similar to what bash version did before.

Made `download` function throw exception if the downloader fails (i.e. curl can't connect).
We are very close to making `downloader.sh` not needed, and be integrated,
but for the moment lets still use it.

Already in Python: ZH, AI, SZ

Closes: https://github.com/openZH/covid_19/issues/152
Closes: https://github.com/openZH/covid_19/issues/182
Closes: https://github.com/openZH/covid_19/issues/184
Closes: https://github.com/openZH/covid_19/issues/201


zdavatz:
good job!

baryluk:
@metaodi PTAL

Issue #206, opened 2020-03-26
track timestamps for failures


zdavatz:
As requested, please track timestamps for failures, see

Use Case: Please always print a timestamp, it does not matter if the scraper fails or not. Just always print a timestamp.

```
~/.software/covid_19/scrapers> ./meta_scrape.sh 
AG 2020-03-25T15:15     319       2 OK 2020-03-26T10:10:21+01:00 https://www.ag.ch/de/themen_1/coronavirus_2/alle_ereignisse/alle_ereignisse_1.jsp
AI 2020-03-28T17:00       9       - OK 2020-03-26T10:10:21+01:00 https://www.ai.ch/themen/gesundheit-alter-und-soziales/gesundheitsfoerderung-und-praevention/uebertragbare-krankheiten/coronavirus
AR 2020-03-26T09:00      40       2 OK 2020-03-26T10:10:21+01:00 https://www.ar.ch/verwaltung/departement-gesundheit-und-soziales/amt-fuer-gesundheit/informationsseite-coronavirus/
BE 2020-03-26T          660       7 OK 2020-03-26T10:10:22+01:00 https://www.besondere-lage.sites.be.ch/besondere-lage_sites/de/index/corona/index.html
BL 2020-03-25T          341       5 OK 2020-03-26T10:10:22+01:00 https://www.statistik.bl.ch/files/sites/Grafiken/COVID19/Grafik_COVID19_BL_Linie.htm
BS 2020-03-25T10:00     466       - OK 2020-03-26T10:10:22+01:00 https://www.gd.bs.ch/, https://www.gd.bs.ch//nm/2020-tagesbulletin-coronavirus-466-bestaetigte-faelle-im-kanton-basel-stadt-gd.html
GE 2020-03-25T12:00    1604      15 OK 2020-03-26T10:10:23+01:00 https://www.ge.ch/document/point-coronavirus-maladie-covid-19/telecharger
GL 2020-03-25T13:30      40       - OK 2020-03-26T10:10:23+01:00 https://www.gl.ch/verwaltung/finanzen-und-gesundheit/gesundheit/coronavirus.html/4817
GR 2020-03-25T          322       6 OK 2020-03-26T10:10:24+01:00 https://www.gr.ch/DE/institutionen/verwaltung/djsg/ga/coronavirus/info/Seiten/Start.aspx
[...]

PR #224, opened 2020-03-26
Add a timestamp to FAILED in meta_scrape.sh

baryluk:
It is not perfect, because it kind of hides the time of data retrival,
but ok.

We should show a bit more failure details in the meta_scrape if possible.

Closes: https://github.com/openZH/covid_19/issues/206


zdavatz:
Thank you!

Issue #205, opened 2020-03-26
Scraper for VD fails when using bash


baryluk:
It works with dash (default `/bin/sh` on Debian/Ubuntu):

```
VD
Scraped at: 2020-03-26T10:05:28+01:00
Date and time: 24.03.2020
Confirmed cases: 2234
Deaths: 21
```

but not with bash

```
VD
Scraped at: 2020-03-26T10:05:32+01:00
Date and time: Date\tHospitalisations
Confirmed cases: l'h\u00f4pital\tD\u00e9c\u00e8s\tTotal
Deaths: de
```

To be investigated.

baryluk:
@zdavatz Out of curiosity which OS, Distro are you using that defaults `/bin/sh` to bash?

baryluk:
BTW. It is probably easier to just port it to Python, like all the other scrapers, and not deal with this dash / bash difference. (it could be a bug in one of the shells tho. who knows).

zdavatz:
> @zdavatz Out of curiosity which OS, Distro are you using that defaults `/bin/sh` to bash?

Gentoo Linux ;)

baryluk:
OK. I was expecting that.

Looks like using `echo -e` will help.

metaodi:
Closing since I merged #236

PR #230, opened 2020-03-26
Use 'echo -e' in the VD scraper so it works with bash

baryluk:
The issue is somewhere before that with '\t', but I don't know where.

Using `echo -e` makes it work fine in both bash and dash.

This is just a temporary solution, before porting this last one to Python.

Closes: https://github.com/openZH/covid_19/issues/205


metaodi:
Closing since I merged #236

PR #235, opened 2020-03-26
Check output of scrapers more carefully, and detect some common failures

baryluk:
Lack of date or cases, is a critical issue, so handle it explicitly.

Example:

```
$ ./scrape_ur.sh | ./parse_scrape_output.py
UR -                     38       - FAILED 2020-03-26T16:41:10+01:00 Missing date https://www.ur.ch/themen/2920
$ ./scrape_sz.sh | ./parse_scrape_output.py
SZ 2020-03-26T            -       - FAILED 2020-03-26T16:41:17+01:00 Missing cases https://www.sz.ch/behoerden/information-medien/medienmitteilungen/coronavirus.html/72-416-412-1379-6948
$
```

Additionally, exit with error.

Output for `OK` is the same.

Make `add_db_entry.py` only accept OK entries,
and exit with error, if some input lines were FAILED.


scrapers/parse_scrape_output.py
```diff
@@ -196,7 +201,15 @@ def parse_date(d):
...
-  print("{:2} {:<16} {:>7} {:>7} OK {} {}".format(abbr, date, cases, deaths if not deaths is None else "-", scrape_time, ", ".join(url_sources)))
+  if date and cases:
+    print("{:2} {:<16} {:>7} {:>7} OK {} {}".format(abbr, date, cases, deaths if not deaths is None else "-", scrape_time, ", ".join(url_sources)))
+  else:
+    errs = []
+    if not date:
+      errs.append("Missing date")
+    if not cases:
+      errs.append("Missing cases")
+    print("{:2} {:<16} {:>7} {:>7} FAILED {} {} {}".format(abbr, date if date else "-", cases if not cases is None else "-", deaths if not deaths is None else "-", scrape_time if not scrape_time is None else "-", ". ".join(errs), ", ".join(url_sources)))
```
metaodi: In the error case, you should raise an Exception so that the `except` block is reached and sys.exit(1) is called.

baryluk: This is exactly what I don't want.

I want to handle the error, print a message, and continue without a program without an error.

metaodi:
Hm, I have to check the workflow then. Because if the error is swallowed here, I have to make sure job fails later on. GitHub Actions fail on sys.exit(1), which is what we want if there was an error.

baryluk:
> Hm, I have to check the workflow then. Because if the error is swallowed here, I have to make sure job fails later on. GitHub Actions fail on sys.exit(1), which is what we want if there was an error.

Ok. Let me actually make do `exit(1)`. Good point.

baryluk:
Please take another look.

baryluk:
Don't you check for `"OK"` in the regexp in `add_db_entry.py` ? If it fails to match, you should throw there too.

metaodi:
Perfect, this makes the whole thing more  stable. Thanks!

PR #62, opened 2020-03-21
Improve schema validation

herrstucki:
- [x] Move `schema.json` to `fallzahlen_kanton_total_csv/`
- [x] Adapt validation script to take directory as argument to allow different schemas per directory
- [x] Add properties from README to schema
- [ ] Fix CSVs that don't pass validation ✌️ 
- [ ] Remove redundant validation script

PR #245, opened 2020-03-26
A workflow to validate the CSVs

metaodi:
This is a minimal approach to validate the CSVs using some regex and checking if all required fields are present.

This PR supersedes #62


metaodi:
This currently fails, because there actually is an invalid CSV in the repo: [COVID19_Fallzahlen_FL_total.csv](https://github.com/openZH/covid_19/blob/master/fallzahlen_kanton_total_csv/COVID19_Fallzahlen_FL_total.csv)

PR #248, opened 2020-03-26
Add scraper workflows for FR, GL, SZ and VD

metaodi:


metaodi:
Wait with merge until #249 is merged.

PR #249, opened 2020-03-26
Fix scraper for new table on FR page

### LINKED ISSUE #255: Python Scraper FR still fails
cc @baryluk 

```
Traceback (most recent call last):
  File "././scrape_fr.sh", line 49, in <module>
    assert m, "Can't find table"
AssertionError: Can't find table
FR -                      -       - FAILED 2020-03-27T09:23:46+01:00 Missing date. Missing cases https://www.fr.ch/covid19/sante/covid-19/coronavirus-statistiques-evolution-de-la-situation-dans-le-canton
```

metaodi:
Closes #255

### DIFF
--- a/scrapers/scrape_fr.sh
+++ b/scrapers/scrape_fr.sh
@@ -66,7 +66,7 @@ d = d.replace('&nbsp;', '')
 """
 
 # For some magical reasons, it works without re.MULTILINE | re.DOTALL.
-r = re.search(r'<tr>\s*<td.*?>(\d\d\.\d\d\.\d\d)</td>\s*<td>([0-9]+| *)</td>\s*<td.*?>([0-9]+| *)</td>\s*<td.*>([0-9]+| *)</td>', d, flags=re.I)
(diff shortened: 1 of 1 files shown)


baryluk:
Could you maybe make the 2 changes:

Use `<t[dh]>` and `</t[dh]>`so it matches both td and th.

Use `<td.*?>` for non greedy match.

scrapers/scrape_fr.sh
```diff
@@ -45,12 +45,12 @@ sc.timestamp()
 </div>
 """
 
-m = re.search(r'<table class="table table-condensed table-hover">(.*?)</table>', d, flags=re.MULTILINE | re.DOTALL)
+m = re.search(r'<table class="table table-striped.*">(.*?)</table>', d, flags=re.MULTILINE | re.DOTALL)
```
baryluk: Lets use:

```
m = re.search(r'<table class="table.*?">(.*?)</table>', d, flags=re.MULTILINE | re.DOTALL)
```

metaodi:
Good point about non-greedy match!

metaodi:
@baryluk PTAL

Issue #250, opened 2020-03-26
REST-API in own repo


apfeuti:
I've created a REST-API to get the data collected by openZH. Data can be filtered by country (CH, FL), cantons and date. Figures for whole Switzerland are also available (totals over all cantons). Output is available in json and csv.

https://covid19-rest.herokuapp.com/
https://github.com/apfeuti/covid19-rest

zukunft:
Thanks. Maybe have also a look here for future ideas: https://covidtracking.com/software/ and https://github.com/COVID19Tracking/covid-tracking-api . This seems to be a similar project.

apfeuti:
Yes, they are similar, but seems to be dedicated to data of the US. There are many COVID19-Rest-APIs out there, most of them are based on data of https://github.com/CSSEGISandData/COVID-19 (Johns Hopkins University).

Maybe openZH can somehow provide their data per canton (area) to JHU? Currently they are reporting for Switzerland only one grand-total.

PR #254, opened 2020-03-27
add link to REST-API in README

apfeuti:
I've created a REST-API to get the data collected by openZH. Data can be filtered by country (CH, FL), cantons and date. Figures for whole Switzerland are also available (totals over all cantons). Output is available in json and csv.


tlorusso:
closes #250

PR #270, opened 2020-03-27
Use empty string instead of None for add_db_entry.py

fabian:
I've noticed there's usually two commits from the scraper:

- one adding the new data: https://github.com/openZH/covid_19/commit/8ed511b1617e5c9d6849f14b2830362d4e9c5897
- a second one adding quotes for empty fields: https://github.com/openZH/covid_19/commit/613737e06c7292ec7c7bc229c6f81e9e9e179886

It looks like when the existing data is loaded from CSV, an empty string is used instead of `null`/`None`. So lets always use an empty string.


metaodi:
Good point, thanks @fabian!

Issue #237, opened 2020-03-26
Expose extra (optional) data in scraper parser output


baryluk:
Right now we only print confirmed cases and deaths (or `-` if not known). But in many scrapers we capture some extra information (like hospitalized, recovered, etc.).

We already are reading it in parser, but not exposing it out.

To not break existing tools, the best is to simply add a "new" parser, or a parser flag, that will then use a new format.

Possible, CSV with headers directly.

metaodi:
Yes I was thinking the same, but I'm not yet sure about the format. Since the result is anyway parsed and put in SQLite currently, headers are not necessary. On the other hand, it would allows us to change the output again in the future.

baryluk:
I just started working on it. I will try to make the output format easy to parse in other tools, but be expandable for our use cases.

baryluk:
Would something like:

```
AG 2020-03-27T16:00     364       2 OK 2020-03-27T16:14:27+01:00 https://www.ag.ch/de/....jsp # Extra: ncumul_hosp=12,ncumul_ICU=10,ncumul_vent=8,ncumul_released=30
```

Maybe work? It would be still human readable, and easy to extract in tools (just some match on `Extra: ([^#]+)` plus `split` on `,` and `=`, and not brake existing tools.

I can update `add_db_entry.py` to work with this too.

baryluk:
BTW. Are ncumul_ICU ncumul_vent actually cumulative, or instantaneous?

baryluk:
Looks promising:

```
$ ./scrape_ge.sh | ./parse_scrape_output.py | SCRAPER_KEY=GE SCRAPER_SOURCE=test ./add_db_entry.py
{'date': '2020-03-27', 'time': '12:00', 'area': 'GE', 'tested': '', 'confirmed': 1924, 'hospitalized': 313, 'icu': 54, 'vent': '', 'released': '', 'deceased': 23, 'source': 'test'}
$ ./scrape_vd.sh | ./parse_scrape_output.py | SCRAPER_KEY=VD SCRAPER_SOURCE=test ./add_db_entry.py 
{'date': '2020-03-26', 'time': '', 'area': 'VD', 'tested': '', 'confirmed': 2532, 'hospitalized': 327, 'icu': '', 'vent': '', 'released': 148, 'deceased': 38, 'source': 'test'}
```

PR #275, opened 2020-03-27
Output extra optional data from parser for scrapers.

baryluk:
Some scrapers already export extra information, like ICU, Vent,
Hospitalized, Recovered.

Export this optional data from parser, and consume in add_db_entry.py

Closes: https://github.com/openZH/covid_19/issues/237


baryluk:
I am open to suggestions before merging this.

metaodi:
LGTM, I will test this for some scrapers tonight and then give feedback or merge.

scrapers/add_db_entry.py
```diff
@@ -50,6 +51,24 @@
         else:
             data['deceased'] = int(data['deceased'])
 
+        # Parse optional data.
+        rest = match.group(5)
+        extras_match = re.search('# Extras: ([^#]+)', rest)
+        if extras_match.group(1):
```
metaodi: This should be `if extras_match:`, because if there was not match, you'll get `Error: 'NoneType' object has no attribute 'group'`

baryluk: Good catch!

baryluk:
> Looks good, just a minor issue.

Fixed.

metaodi:
Thanks @baryluk!

PR #277, opened 2020-03-27
adding VS hospitalizations

zukunft:
In today's report the hospitalizations of the past days have also been adjusted. Should we include these updates?


zukunft:
The question remains, if the history should be adjusted. See #278

PR #280, opened 2020-03-27
Don't use --no-progress-meter with curl in download.sh

baryluk:
It looks like old versions (before 7.67) of curl doesn't support it.


metaodi:
Are you omitting --silent on purpose?

baryluk:
Oh. My bad.

It is fine for it to be without --silent, adjusted the PR message and commit message.

Issue #279, opened 2020-03-27
Python Scraper GR failed


zdavatz:
```
GR 2020-03-27T            -       - FAILED 2020-03-27T22:30:05+01:00 https://www.gr.ch/DE/institutionen/verwaltung/djsg/ga/coronavirus/info/Seiten/Start.aspx # Errors: Missing cases. Confirmed cases is None. Deaths is None
```
cc @baryluk

baryluk:
#zdavatz Also, can you stop saying "Python Scraper" in the issue titles. It is just scraper. It is hugely irrelevant that it is or not a Python.

Issue #271, opened 2020-03-27
New data source (JSON) for GR


andreasamsler:
https://www.gr.ch/DE/institutionen/verwaltung/djsg/ga/coronavirus/_layouts/15/GenericDataFeed/feed.aspx?PageID=26&ID=g_1175d522_e609_4287_93af_d14c9efd5218&FORMAT=JSONRAW

baryluk:
Good. Will consume it.

Could you also provide URL for a page that uses it normally on www.gr.ch?

andreasamsler:
By tomorrow GR is going to switch https://www.gr.ch/DE/institutionen/verwaltung/djsg/ga/coronavirus/info/Seiten/Start.aspx to consume their JSON.

baryluk:
I have a working prototype:

```
$ ./scrape_gr.sh 
GR
Downloading: https://www.gr.ch/DE/institutionen/verwaltung/djsg/ga/coronavirus/info/Seiten/Start.aspx
Scraped at: 2020-03-28T01:31:46+01:00
Date and time: 27.03.2020
Confirmed cases: 409
Deaths: 9
Downloading: https://www.gr.ch/DE/institutionen/verwaltung/djsg/ga/coronavirus/_layouts/15/GenericDataFeed/feed.aspx?PageID=26&ID=g_1175d522_e609_4287_93af_d14c9efd5218&FORMAT=JSONRAW
Scraped at: 2020-03-28T01:31:46+01:00
Date and time: 2020-03-27
Confirmed cases: 409
Hospitalized: 52
Deaths: 9
$
```

```
$ ./scrape_gr.sh | ./parse_scrape_output.py
GR 2020-03-27           409       9 OK 2020-03-28T01:32:27+01:00 # Extras: ncumul_hosp=52 # URLs: https://www.gr.ch/DE/institutionen/verwaltung/djsg/ga/coronavirus/info/Seiten/Start.aspx, https://www.gr.ch/DE/institutionen/verwaltung/djsg/ga/coronavirus/_layouts/15/GenericDataFeed/feed.aspx?PageID=26&ID=g_1175d522_e609_4287_93af_d14c9efd5218&FORMAT=JSONRAW
```

It looks nice and simple, and should be robust, and not fail even if some fields are missing, or the URL goes away, then it reverses to the previous method.

PR #281, opened 2020-03-27
Fix GR scraper

baryluk:
It got '&#58;' (ASCII 58 is ':') instead of literal ':'.
Just replace it back.

Add an example.

Closes: https://github.com/openZH/covid_19/issues/279


metaodi:
What about #271? Consuming the JSON should be super easy. Or do you want to fix this first?

baryluk:
This first. This is easy fix.

Consuming JSON next, it is easy, but still non-zero amount of work :)

metaodi:
Gottcha!

baryluk:
@metaodi Thank you!

PR #286, opened 2020-03-28
Add a new source for GR

baryluk:
Uses JSON data, which we can parse directly.

The table is sorted, but, to be sure, sort it anyway.

It looks to be using our CSV formatting essentially.

Closes: https://github.com/openZH/covid_19/issues/271

### DIFF
--- a/scrapers/parse_scrape_output.py
+++ b/scrapers/parse_scrape_output.py
@@ -135,7 +135,7 @@ def parse_date(d):
     assert 1 <= int(mo[3]) <= 23
     # 24.3. / 10h
     return f"2020-{int(mo[2]):02d}-{int(mo[1]):02d}T{int(mo[3]):02d}:00"
-  mo = re.search(r'^(\d\d\d\d-\d\d-\d\d)T?(\d\d:\d\d)(:\d\d)?$', d)
+  mo = re.search(r'^(\d\d\d\d-\d\d-\d\d)[ T](\d\d:\d\d)(:\d\d)?$', d)
   if mo:
     # 2020-03-23T15:00:00
     # 2020-03-23 15:00:00
--- a/scrapers/scrape_gr.sh
+++ b/scrapers/scrape_gr.sh
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 
+import json
 import scrape_common as sc
 
 print('GR')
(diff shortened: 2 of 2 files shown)


andreasamsler:
@metaodi thanks for reviewing and merging

metaodi:
@andreasamsler this PR is not yet ready for review (it's a draft). I'll take care once it's ready.

baryluk:
PTAL

metaodi:
@baryluk is there a reasin for keeping the "old" source for the scraper? So now the scraper has two sources and potentially outputs data from both (?).
How does the parser handle this? I'm currently on mobile so can't test myself.

baryluk:
> @baryluk is there a reasin for keeping the "old" source for the scraper? So now the scraper has two sources and potentially outputs data from both (?).
> How does the parser handle this? I'm currently on mobile so can't test myself.

Parser will use last non-None / parsable emitted entry from the scraper. With exception of a date, where it will use more accurate source (the one with day and time, not just time). 

The reason? I don't trust one or another, so wanted to have both, and failure of first or second one, to be not critical. If the first fails, it will still use data from the second one. And vice versa:

Working:

```
$ ./scrape_gr.sh 
GR
Downloading: https://www.gr.ch/DE/institutionen/verwaltung/djsg/ga/coronavirus/info/Seiten/Start.aspx
Scraped at: 2020-03-28T19:07:07+01:00
Date and time: 27.03.2020
Confirmed cases: 409
Hospitalized: 52
Deaths: 9
Downloading: https://www.gr.ch/DE/institutionen/verwaltung/djsg/ga/coronavirus/_layouts/15/GenericDataFeed/feed.aspx?PageID=26&ID=g_1175d522_e609_4287_93af_d14c9efd5218&FORMAT=JSONRAW
Scraped at: 2020-03-28T19:07:07+01:00
Date and time: 2020-03-27
Confirmed cases: 409
Hospitalized: 52
Deaths: 9
$ ./scrape_gr.sh  | ./parse_scrape_output.py 
GR 2020-03-27           409       9 OK 2020-03-28T19:09:26+01:00 # Extras: ncumul_hosp=52 # URLs: https://www.gr.ch/DE/institutionen/verwaltung/djsg/ga/coronavirus/info/Seiten/Start.aspx, https://www.gr.ch/DE/institutionen/verwaltung/djsg/ga/coronavirus/_layouts/15/GenericDataFeed/feed.aspx?PageID=26&ID=g_1175d522_e609_4287_93af_d14c9efd5218&FORMAT=JSONRAW
$
```

But, I think you are right, we can probably remove the first (old) source, and use the JSON only.  If it breaks we can restore the old source. The issue I have with that, is that the old source uses a user visible website, which is likely to be updated, but the JSON source could become stale / abandoned.

We don't really loose much by having both in the scraper. Parser will extract as much as it can from one or another, or both.

metaodi:
I think it's fine. I just remember someone mentioning that GR actually uses the JSON to feed their own website (didn't check myself). If it's actually true, it's quite pointless to scrape both. But until we know for sure,  let's keep both.

baryluk:
@metaodi You are probably right. Thanks for merging. I will work on it to maybe only fallback to an old source, if first fails to load or don't have most recent data.

PR #347, opened 2020-03-31
Add source parser for AG

fabian:
- Add support for `source` as an extra parameter
- Scrape source for AG
- Update todays source for AG

### DIFF
--- a/scrapers/scrape_ag.sh
+++ b/scrapers/scrape_ag.sh
@@ -23,3 +23,5 @@ print('ICU:', sc.find(r'([0-9]+) Person(en)? werden auf Intensivstationen behand
 print('Vent:', sc.find(r'([0-9]+) Person(en)? k(ü|&uuml;)nstlich beatmet werden', d))
 
 print('Deaths:', sc.find(r'([0-9]+) Person(en)? an den Folgen des Coronavirus ?verstorben', d))
+
+print('Downloading:', 'https://www.ag.ch' + sc.find(r'<a .*?href="(.+?)".*?>.*?Lagebulletin.*?</a>', d))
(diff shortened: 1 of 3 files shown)


fabian:
Might conflict with https://github.com/openZH/covid_19/pull/348 - will update once that has been merged. 😉

metaodi:
So I will merge then. I think I should cover this case already, but let's see.

metaodi:
Okay probably not, since we don't actually scrape the PDF, the link is currently missing in the output. Would parsing the PDF be an option?

fabian:
It would be an option, but I think it's easier to just parse the HTML and then use the PDF as a reference source.

scrapers/scrape_ag.sh
```diff
@@ -23,3 +23,5 @@ print('ICU:', sc.find(r'([0-9]+) Person(en)? werden auf Intensivstationen behand
 print('Vent:', sc.find(r'([0-9]+) Person(en)? k(ü|&uuml;)nstlich beatmet werden', d))
 
 print('Deaths:', sc.find(r'([0-9]+) Person(en)? an den Folgen des Coronavirus ?verstorben', d))
+
+print('Source:', 'https://www.ag.ch' + sc.find(r'<a .*?href="(.+?)".*?>.*?Lagebulletin.*?</a>', d))
```
metaodi: We already have everything in place to gather URLs that are being used in the scraper, instead of introducing a new key `Source:` that needs extra parsing, I'd suggest to change this line to

```
print('Downloading:', 'https://www.ag.ch' + sc.find(r'<a .*?href="(.+?)".*?>.*?Lagebulletin.*?</a>', d))
```

And everything else is already taken care of in the parser.

fabian: Okay, makes sense - updated!

andreasamsler:
Great work, friends!

PR #348, opened 2020-03-31
Fix BL scraper

metaodi:


baryluk:
Nice trick with using iframe to get latest graph. I was worried the URL will become stale. Awesome.

PR #368, opened 2020-04-01
Fix by adding a missing ICU column extraction

piotrpiotrowski:


baryluk:
Great fix. Thank you.

PR #369, opened 2020-04-01
Revert manual intervention

piotrpiotrowski:


metaodi:
What is this PR about?

metaodi:
ah I see now. Thanks!

piotrpiotrowski:
I fixed VD scraper and it was run successfully adding the last line to `csv` with a date 2020.03.31.
Meanwhile the csv was updated manually with a date 2020.04.01, both lines has the same values.
I think the manual adjustment should be deleted, because scraper will collect latest data next time it will be triggered.

PR #377, opened 2020-04-01
Fix BL scraper

baryluk:
New column in the table

### DIFF
--- a/scrapers/scrape_bl.sh
+++ b/scrapers/scrape_bl.sh
@@ -33,13 +33,26 @@ Datum, Bestätigte Fälle, Verstorbene
 </pre>
 """
 
+# 2020-04-01
+"""
+<pre id="data" style="display:none;">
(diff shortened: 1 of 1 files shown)


baryluk:
Is `Geheilte geschätzt` recovered? I am not sure.

metaodi:
Yes it's recovered. It's an assumption, but I guess it's like that in all cantons, no one really measures the recovered cases. [There is a description of this concept from BL](https://www.baselland.ch/politik-und-behorden/direktionen/volkswirtschafts-und-gesundheitsdirektion/amt-fur-gesundheit/medizinische-dienste/kantonsarztlicher-dienst/aktuelles/covid-19-faelle-kanton-basel-landschaft/pdf/konzept-geheilte-bl.pdf/@@download/file/Konzept_Geheilte_BL.pdf).

baryluk:
Please take another look.

PR #378, opened 2020-04-01
Re-write LU scraper with beautifulsoup

metaodi:
Parsing tables is a pain, so I decided to give it another try using beautifulsoup.


metaodi:
@baryluk would like to have your feedback here.

baryluk:
@metaodi Looks good to me.

PR #384, opened 2020-04-02
fixing sg scraper

dominikgehl:


jb3-2:
@dominikgehl Scraper currently errors out, any idea why? Thanks for your help. 

See https://github.com/openZH/covid_19/runs/555010371?check_suite_focus=true

Run python scrapers/scraper_canton_sg.py
  python scrapers/scraper_canton_sg.py
  shell: /bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.7.6/x64
Error: 'NoneType' object has no attribute 'parent'
Traceback (most recent call last):
  File "scrapers/scraper_canton_sg.py", line 106, in <module>
    parse_page(soup, conn)
  File "scrapers/scraper_canton_sg.py", line 36, in parse_page
    boxes = soup.find("h3", string=re.compile("Update Kanton St.Gallen")).parent.find_all("p")
AttributeError: 'NoneType' object has no attribute 'parent'

dominikgehl:
> @dominikgehl Scraper currently errors out, any idea why? Thanks for your help.
> 
> See https://github.com/openZH/covid_19/runs/555010371?check_suite_focus=true
> 
> Run python scrapers/scraper_canton_sg.py
> python scrapers/scraper_canton_sg.py
> shell: /bin/bash -e {0}
> env:
> pythonLocation: /opt/hostedtoolcache/Python/3.7.6/x64
> Error: 'NoneType' object has no attribute 'parent'
> Traceback (most recent call last):
> File "scrapers/scraper_canton_sg.py", line 106, in
> parse_page(soup, conn)
> File "scrapers/scraper_canton_sg.py", line 36, in parse_page
> boxes = soup.find("h3", string=re.compile("Update Kanton St.Gallen")).parent.find_all("p")
> AttributeError: 'NoneType' object has no attribute 'parent'

From the error message it seems that soup.find("h3", string=re.compile("Update Kanton St.Gallen")) returns None

dominikgehl:
Ok. I can reproduce the issue. I'll have a look. But why are there actually these two files scrape_sg.sh and scraper_canton_sg.py ?

dominikgehl:
Ok, so this error can be fixed by using
boxes = soup.find("h3", string=re.compile("Update ganzer Kanton St\.Gallen")).parent.find_all("p")
But everything else would also need to be fixed because the whole HTML page setup was changed. Which is what I tried to solve in scrape_sg.sh ...

jb3-2:
Thanks! That seems to be a legacy, I think @metaodi has just fixed this in https://github.com/openZH/covid_19/commit/9b837b84fc5a7bd6675cf84e75d16f76a3d60210

metaodi:
Wait! This is the "old" scraper, appearently it was still setup in the GitHub workflow. I just updated the workflow to use `scrape_sg.sh`. So let's see when it runs the next time.

Sorry about that!

PR #420, opened 2020-04-03
Fix ICU for scraper AG

fabian:
Slight change in the wording on ag.ch caused a wrong number. Fixed the scraper and the CSV.


baryluk:
LGTM

PR #423, opened 2020-04-04
fixing hospitalized number for Glarus

dominikgehl:


scrapers/scrape_gl.sh
```diff
@@ -36,5 +36,5 @@ d = d.replace('<strong>', '').replace('</strong>', '')
 
 print('Date and time:', sc.find(r'Update\s*(.+ Uhr)\)<', d))
 print('Confirmed cases:', sc.find(r'Bestätigte\s*Fälle\s*:\s*([0-9]+)\b', d))
-print('Hospitalized:', sc.find(r'Hospitalisierungen\s*:\s*([0-9]+)\b', d))
+print('Hospitalized:', sc.find(r'Personen in Spitalpflege\s*:\s*([0-9]+)\b', d))
```
metaodi, line 39: Please make sure the old and new format is matched. More often than not, these things are switched back and forth. So use something like `(?:Hospitalisierungen|Personen in Spitalpflege)`

dominikgehl: Thanks for the suggestion. PR is updated.

PR #425, opened 2020-04-04
fixing French calendar month names

dominikgehl:


scrapers/parse_scrape_output.py
```diff
@@ -24,17 +24,17 @@
 
 months_fr = {
   "janvier": 1,
-  "fèvrier": 2,
+  "février": 2,
```
metaodi, line 27: Since we use this for parsing, let's just have both variants in the dict. So if it's different on the website, we can still parse it correctly.

scrapers/parse_scrape_output.py
```diff
@@ -24,17 +24,17 @@
...
   "mars": 3,
   "avril": 4,
   "mai": 5,
   "juin": 6,
   "juillet": 7,
-  "aout": 8,
+  "août": 8,
```
metaodi, line 33: dito

scrapers/parse_scrape_output.py
```diff
@@ -24,17 +24,17 @@
...
   "septembre": 9,
   "octobre": 10,
   "novembre": 11,
-  "decembre": 12,
+  "décembre": 12,
```
metaodi, line 37: dito

scrapers/parse_scrape_output.py
dominikgehl: Updated all 3

metaodi:
Thank you @dominikgehl

Issue #431, opened 2020-04-04
LU scraper failed


metaodi:
[LU scraper currently fails](https://github.com/openZH/covid_19/runs/560412179?check_suite_focus=true):

```
Run the scraper...
./scrapers/run_scraper.sh: line 28: /home/runner/work/covid_19/covid_19/scrapers/scrape_lu.py: No such file or directory
Error: Not matched input line: ?? Error: unsupported format string passed to NoneType.__format__
Error: Not matched input line: Traceback (most recent call last):
Error: Not matched input line: File "/home/runner/work/covid_19/covid_19/scrapers/parse_scrape_output.py", line 304, in <module>
Error: Not matched input line: f"
# Errors: {errs}"))
Error: Not matched input line: TypeError: unsupported format string passed to NoneType.__format__
Error: Not matched input line: 
```

baryluk:
Yes, one of the recent commits renamed it to `scrape_ly.py` apparently. Should be `scrape_lu.py`

PR #433, opened 2020-04-04
Rename scrape_ly.py to scrape_lu.py

baryluk:
Looks like a bad change in https://github.com/openZH/covid_19/pull/428

Closes: https://github.com/openZH/covid_19/issues/431


metaodi:
Oh I missed that, good catch!

Issue #68, opened 2020-03-21
better monitoring of the updating process


tlorusso:
As more and more people is involved (yay!) in updating the data and building scrapers, we might have to start thinking about how to monitor the updating process, and think about how we can ensure that the data is being updated even if one of the scrapers fails or someone forgets to check for new data.

Do you habe suggestions how we could manage this @metaodi @baryluk @ebeusch @herrstucki @andreasamsler @zdavatz ?

A table in the Readme (or somewhere else?) which is refreshed automatically after each push to a single file might help, similar to what @baryluk has created here:

https://github.com/openZH/covid_19/issues/61

The table we have now is built by hand.

baryluk:
That is a good point. Would be nice to know which stuff needs attention or is outdated.

ebeusch:
I think at the moment we lack a good overview of which cantons are outdated (incl. historic data) and where it might be necessary to get information by going through news outlets etc.

At the moment there also seems to be some double work. This includes both scrapers and manual updates. So, once there are scrapers (@baryluk seems to get one out by the minute!) it might also be good to assign who is running the scrapers (+ second/third responsible in case people fall ill/are prevented from updating. Similarly, teams can be assigned to update cantons where it's not possible to have automatic data collection.

rokroskar:
The covidtracking project (covidtracking.com) uses this: https://github.com/COVID19Tracking/covid-tracking - it generates diffs of the sites they are monitoring for updates and shoots off a message to slack when one needs attention. Seems like something similar might be useful here? I think here it would be great if the scrapers could be consolidated into one codebase so that the error reporting etc. can be a bit more standardized.

baryluk:
Yeah. I did not expect to write so many, but they all work. I will put them into my fork of the repo, and add a script to run all of them, normalize various date formats, and create a consolidated output of new data. Then we could run it from cron, or manually. Once I have few running, I will open MR.

zdavatz:
very good work @baryluk !!!!

metaodi:
@baryluk I will work on consolidating the scrapers tonight, maybe we can exchange later on mattermost to join forces. I want to schedule all scrapers with Github actions.

zdavatz:
I would go the way @baryluk suggests. 

I suggest you have cronjob that runs his scrappers after the media conferences of the cantons or lets say every two hours. If the scrapper finds new data, then github actions should push the new numbers to github, so that data can be visualized.

[...]

PR #439, opened 2020-04-04
Add a matrix of expectations for each scraper in the parser.

baryluk:
This is an independent list that is checked against what scraper
provides. If any expected number is missing, it is reported as warning to
stderr.

In the future we might turn it into an error and fail parsing, to
indicate stronger that the scraper is broken or something.

This is related to https://github.com/openZH/covid_19/issues/68
and https://github.com/openZH/covid_19/issues/68#issuecomment-605527035
in particular.

### DIFF
--- a/scrapers/parse_scrape_output.py
+++ b/scrapers/parse_scrape_output.py
@@ -198,6 +198,7 @@ def maybe_new_int(name, value, old_value, required=False):
             warns.append(f"{name} ({value}) not a number")
     return old_value
 
+import scrape_matrix as sm
 
 try:
     i = 0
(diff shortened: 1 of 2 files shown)


baryluk:
Can be probably merged as is, as it works for me, and provides some good reminder in central place, but I am open to any comments about it.

metaodi:
I really like it. I was just wondering two things:

1. I think a run should fail if a field is missing, otherwise we'll miss these warnings
2. May you should move the the check for the confirmed cases and the date to the matrix as well. 

But of course we can start with this and extend it later.

baryluk:
Lets start with this. Thanks for the comments, I agree they should implemented. Will try to do this tomorrow.

PR #440, opened 2020-04-05
Adds NLP scraper to scraper_bs.sh

abieler:
Updated `scrape_bs.sh` to run with the new nlp model.
added a `utils` module to keep all nlp stuff in `utils/bs`.


scrapers/scrape_bs.sh
```diff
@@ -14,14 +33,18 @@ URL = sc.filter(r'Tagesbulletin.*Corona', URL)
...
 sc.timestamp()
 
-d = d.replace('&auml;', 'ä')
-d = d.replace('&nbsp;', ' ')
-d = sc.filter(r'positive\s*Fälle', d)
+ptext = bs.preprocess(text)
+result = bs.parse(ptext)
+result = bs.update_indirect_numbers(result)
+
+for k, v in result.items():
```
abieler, line 44: This should probably adapted such that it only spits out whatever you need

scrapers/scrape_bs.sh
```diff
@@ -1,6 +1,25 @@
...
+import utils.bs as bs
+
+def text_from_url(url: str) -> str:
+    req = requests.get(url)
+    html = BeautifulSoup(req.content, "html.parser")
+
+    # I dont really know what I am doing here...
+    # But seems to be working.
+    # The goal is to extract the main text from the URL.
+    texts = []
```
abieler, line 16: i really dont know my way around with html parsing and such.. so this part of the code is surely of low qualitty...

jb3-2:
Thank you very much for your efforts, @abieler. @metaodi Can you find time to review and test this somehow?

metaodi:
I can, but it's still WIP, right @abieler? Or do you already want feedback?

abieler:
@metaodi you can shoot with feedback right away

baryluk:
Could you wrap the call to NLP model in `scrape_bs.sh`, into exception handling block. Most notably, I can't really test it, because I don't have access to `spacy` on my distro (i know it can be installed via pip).

baryluk:
Also, please make sure that: `scrape_bs.py | ./parse_scrape_output.py` does not fail.

abieler:
> Also, please make sure that: `scrape_bs.py | ./parse_scrape_output.py` does not fail.

I guess this was just a typo from you, but I was thinking about renaming `scrape_bs.sh` to `scrape_bs.py` as it really is a python and not a shell script. ok to do so?

baryluk:
> > Also, please make sure that: `scrape_bs.py | ./parse_scrape_output.py` does not fail.
> 
> I guess this was just a typo from you, but I was thinking about renaming `scrape_bs.sh` to `scrape_bs.py` as it really is a python and not a shell script. ok to do so?

We did rename this files few days ago. It is now `scrape_bs.py`. You might need to rebase.

abieler:
@baryluk thanks for all the input. I addressed all of it. please have another look now.

scrapers/scrape_bs.py
```diff
@@ -1,8 +1,28 @@
...
+    # The goal is to extract the main text from the URL.
+    texts = []
+    for p in html.find_all("p")[1:-8]:
+        text = p.get_text()
+        if len(text.split(" ")) > 5:
+            texts.append(text)
+    return " ".join(texts)
 
 print('BS')
+sc.timestamp()
```
baryluk, line 24: Put (move) this `sc.timestamp()` just after `text = text_from_url(f'https://www.gd.bs.ch/{URL}')` on line 38.

scrapers/scrape_bs.py
```diff
@@ -1,8 +1,28 @@
 #!/usr/bin/env python3
+import requests
+from bs4 import BeautifulSoup
+
 
 import scrape_common as sc
+import utils.bs as bs
+
+def text_from_url(url: str) -> str:
+    req = requests.get(url)
```
baryluk, line 10: Please add this:

```
  print(f'Downloading: {url}')
```

just before `requests.get`, so we capture it in parser and CSV files.

scrapers/scrape_bs.py
```diff
@@ -61,9 +80,9 @@
...
-print('Confirmed cases:', sc.find(r'(?:insgesamt\s*)?([0-9]+)\s*positive', d))
-print('Recovered:', sc.find(r'([0-9]+) Personen der [0-9]+ positiv Getesteten .+ sind wieder genesen', d))
-print('Hospitalized:', sc.find(r'Aktuell befinden sich ([0-9]+) Einwohnerinnen und Einwohner des Kantons Basel-Stadt aufgrund einer Covid-19-Infektion in Spitalpflege', d))
-print('ICU:', sc.find(r'Insgesamt ([0-9]+) Personen benötigen Intensivpflege', d))
+print('Date and time:', sc.find(r'Stand\s*[A-Za-z]*,?\s*(.+?),\s*(?:liegen\s*)?insgesamt', text))
+print("Confirmed cases: ", result["NUMCUL_CONF_RESIDENTS"])
+print("Death: ", result["NUMCUL_DECEASED"])
+print("Recovered: ", result["NUMCUL_RELEASED"])
+print("Hospitalized: ", result["NUMCUL_HOSP"])
+print("ICU: ", result["NUMCUL_ICU"])
```
baryluk: Minor thing, but please remove the trailing spaces after a colon (print automatically puts a separator, default space, between printed elements), and use single quotes, to be consistent.

Example:

```python3
print('Confirmed cases:', result["NUMCUL_CONF_RESIDENTS"])
```
Same for other print statements here.

Thanks.

scrapers/utils/bs/bs.py
```diff
@@ -0,0 +1,52 @@
+import argparse
+import logging
+
+logging.basicConfig(level=logging.ERROR)
+
+
+try:
```
baryluk, line 7: I don't think you need `try` / `except` here. Just do `import spacy`.

scrapers/scrape_bs.py
```diff
@@ -14,14 +34,16 @@
...
-URL = sc.filter(r'href', URL)
-URL = URL.split('"')[1]
-d = sc.download(f'https://www.gd.bs.ch/{URL}')
-sc.timestamp()
+URL = sc.filter(r'href', URL).split('"')[1]
+text = text_from_url(f'https://www.gd.bs.ch/{URL}')
+
+ptext = bs.preprocess(text)
+try:
+    result = bs.parse(ptext)
```
baryluk, line 42: You can also move the `import utils.bs as bs` under this `try` block?

scrapers/utils/bs/bs.py
abieler: i was thinking about your earlier comment for cases where you dont have spacy installed. 
sure i should remove?

abieler:
@baryluk next try :)

scrapers/scrape_bs.py
```diff
@@ -61,9 +84,10 @@
...
-print('Recovered:', sc.find(r'([0-9]+) Personen der [0-9]+ positiv Getesteten .+ sind wieder genesen', d))
-print('Hospitalized:', sc.find(r'Aktuell befinden sich ([0-9]+) Einwohnerinnen und Einwohner des Kantons Basel-Stadt aufgrund einer Covid-19-Infektion in Spitalpflege', d))
-print('ICU:', sc.find(r'Insgesamt ([0-9]+) Personen benötigen Intensivpflege', d))
+
+print('Date and time:', sc.find(r'Stand\s*[A-Za-z]*,?\s*(.+?),\s*(?:liegen\s*)?insgesamt', text))
+print("Confirmed cases:", result.get("NUMCUL_CONF_RESIDENTS", ""))
+print("Death:", result.get("NUMCUL_DECEASED", ""))
+print("Recovered:", result.get("NUMCUL_RELEASED", ""))
+print("Hospitalized:", result.get("NUMCUL_HOSP", ""))
+print("ICU:", result.get("NUMCUL_ICU", ""))
```
baryluk, line 93: Missing new line at the end of the file.

baryluk:
LGTM in general. Worth a try.

scrapers/utils/bs/bs.py
```diff
@@ -0,0 +1,47 @@
...
+        for ent in sdoc.ents:
+            try:
+                result[ent.label_] = int(ent.text.strip('.,'))
+            except ValueError:
+                logging.warning(f"Error parsing: {ent.text.strip('.,')}")
+    return result
+
+
+def update_indirect_numbers(result: dict) -> dict:
+    if result.get("NUMCUL_CONF", None) and result.get("NUMCUL_CONF_RESIDENTS"):
```
baryluk, line 42: You don't need to pass `None` explicitly, it is a default for the `default` argument:

Help on built-in function get:

```
get(key, default=None, /) method of builtins.dict instance
    Return the value for key if key is in the dictionary, else default.
```

baryluk:
@metaodi  I think we can merge it as is, and see how it works in the wild, and improve later. It works nice for me.

scrapers/utils/bs/bs.py
abieler:
```python
if "NUMCUL_CONF" in result and "NUMCUL_CONF_RESIDENTS" in result:
```
nice! i did not know that!

abieler: let me add that, its way more readable than my mess :)

metaodi: Actually in python you would generally not use something like `if "NUMCUL_CONF" in result`, it follows the EAFP coding style.
But I agree: passing None does not make sense.

baryluk:
I guess, we can hold this solution off for the moment, as the simpler code works for the moment.

@abieler How smart is this code to cope with minor text variations? Did you test it by introducing various changes to the text, especially one close to the extracted numbers?

jb3-2:
Agreed - let‘s see how the existing scraper works today. I will have to restrain myself from adding the data manually... ;-)

jb3-2:
The existing scraper works well enough. Let's archive this PR for the moment, if you agree. Thanks for your outstanding work! I have learned a lot from it.

PR #447, opened 2020-04-05
Fix SG scraper. Relax more regexps.

baryluk:
There was a minor change in the table format, that got some colspans on
TD tags. Skip them if any.

Replace all matches on space by `\s*` instead, just in case.

Adjust regexp for hospitalized (isolated) and ICU to match new
abbreviation.

Make date regexp more forgiving, to also match future months.


baryluk:
The failure to extract some fields was detected by expectation matrix code from https://github.com/openZH/covid_19/pull/439

Yay.

PR #461, opened 2020-04-07
Data structure enhancements COVID19_Fallzahlen_Beispiel.csv

andreasamsler:
Enhancements based on developements and discussions by the core team; @tlorusso @mmznrSTAT please review, and propose furthet enhancements, resp. raise constraints, thanks!


metaodi:
Feedback:

- Why is there new_hosp? Is this something that is reported? Is this an important number?
- If so, are other new_* columns needed as well? like new_ICU?
- Is there a reason why ICU is capitalized? I propose to use lowercase for all columns
- What is source_description? Is this something the cantons provide? Or a defined vocabulary?

tlorusso:
i support lowercase for all. The 'new_hops' column adresses a specific but important need to be able to calculate the total number of people that have been hospitalized so far (which is not directly possible with the current_hosp-variable). I guess no Canton reports this number in this way yet, but we have an important request that might justify that we 'encourage' to report this number in this way.

mmznrSTAT:
current_hosp could be understood as currently in hospital on a given date. new_hosp would be something different and could help modelling.

mmznrSTAT:
source_description -> better source_comments

simgraworldwide:
Could you change the readme as well so that we are all definitely on the same page when it comes to what will actually be reported and published differently?

andreasamsler:
Done:
- Reviewed with Update Roundtrip Reviewers, @tlorusso, @mmznrSTAT and @metaodi 
- New example file adapted accordingly

Open:
- New example file to be finally accepted, and introduced by Thursday morning

metaodi:
I'm currently working on a script to transform the old files to the new structure and vice-versa.

metaodi:
Here is the PR for the transformation: #472

PR #464, opened 2020-04-07
Extracting Hospitalized and ICU for BL

dominikgehl:


scrapers/scrape_bl.py
```diff
@@ -11,14 +12,15 @@
 <iframe width="100%" class="iframeblock loading" onload="onIframeLoaded(this)" src="https://www.statistik.bl.ch/files/sites/Grafiken/COVID19/20200331_COVID19_BL.htm" scrolling="auto" height="600"></iframe>
 """
 
-iframe = sc.filter(r'<iframe', main_site)
-iframe_url = sc.find(r'src="(.+?)"', iframe)
+soup = BeautifulSoup(main_site, 'html.parser')
+for iframe in soup.find_all('iframe'):
```
metaodi, line 16: Do you really need multiple iframes? If you use `soup.find('iframe')`, you'll get the first iframe only.

dominikgehl: Yes, but unfortunately, the hospitalization data is in the 2nd iframe ... we need to fetch them both.

Issue #357, opened 2020-03-31
NE scraper fails


zdavatz:
```
NE -                      -       - FAILED 2020-03-31T20:00:07+02:00  # URLs: https://www.ne.ch/autorites/DFS/SCSP/medecin-cantonal/maladies-vaccinations/Pages/Coronavirus.aspx # Errors: Missing date. Missing cases. Date and time is None. Confirmed cases is None. Deaths is None
```
cc @baryluk

baryluk:
I don't see any numbers any more on https://www.ne.ch/autorites/DFS/SCSP/medecin-cantonal/maladies-vaccinations/Pages/Coronavirus.aspx

I send them an email.

Could you look around yourself, maybe I missed some page where this data is available?

If not send them an email too please.

baryluk:
I see some data in https://www.ne.ch/autorites/DFS/SCSP/medecin-cantonal/maladies-vaccinations/Documents/Covid-19-Statistiques/COVID19_PublicationInternet.pdf

Scrape possible, but still non trivial.

baryluk:
I looked into `pdftotext` for this PDF, and I would say, it is essentially impossible to parse reliably, due to nature of the graph, rotated text, invisible line breaks, and other factors (like there is no guarantees of the data order, and correspondence of data and number of cases).

metaodi:
I checked archive.org, it seems as if they just removed the block with the numbers, that was previously in the section titled "Quelle est la situation actuelle ?"

zdavatz:
@baryluk I think you should be able to grab the number here with "pdfgrep". See:

<img width="1405" alt="Bildschirmfoto 2020-03-31 um 22 04 13" src="https://user-images.githubusercontent.com/4953/78070195-abf05180-739b-11ea-8615-b81b7e523de7.png">

If you know what the last number is ;). Just always take the last number in the row. This is actually Text in the file ;).

baryluk:
@zdavatz That is not usable.

baryluk:
@metaodi Yes,

http://web.archive.org/web/20200329210634/https://www.ne.ch/autorites/DFS/SCSP/medecin-cantonal/maladies-vaccinations/Pages/Coronavirus.aspx

![Screenshot at 2020-03-31 20-22-02](https://user-images.githubusercontent.com/110322/78071645-6462c900-738d-11ea-868f-84326eebf617.png)

baryluk:
[...]

PR #470, opened 2020-04-08
Scrape PDF from NE and extract all numbers

baryluk:
Releated to: https://github.com/openZH/covid_19/issues/357

### DIFF
--- a/scrapers/parse_scrape_output.py
+++ b/scrapers/parse_scrape_output.py
@@ -123,9 +123,10 @@ def parse_date(d):
         assert 2020 <= int(mo[3]) <= 2021
         assert 1 <= int(mo[4]) <= 23
         return f"{int(mo[3]):4d}-{months_all[mo[2]]:02d}-{int(mo[1]):02d}T{int(mo[4]):02d}:00"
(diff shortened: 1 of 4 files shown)


metaodi:
Is this stable? I will try test this morning to get a feeling. All these replacements of arbitrary amounts of whitespace seems to break as soon as the PDF changes a bit.

baryluk:
:man_shrugging: 

It did work fine yesterday too. The random amount of whitespaces there are kind of on purpose, because of how `pdftotext -layout` creates a table format.

I should make a bit better comment about it in the source probably.

baryluk:
@maekke97  What do you think?

baryluk:
I added some comments to the scraper, ( https://github.com/openZH/covid_19/blob/f60135e0ed6df5946e2c0b39de9948487489bdf2/scrapers/scrape_ne.py ), that should explain the parsing a bit better.

maekke97:
@baryluk looks good to me except that `d` gets assigned a static string at the moment instead of the one from the pdf. My other comments are rather optional.

baryluk:
@maekke97 Thanks for the review. Addressed the issues.

metaodi:
@baryluk @maekke97 Thank you to both of you!

PR #479, opened 2020-04-08
Fix a date regexp regression / break due to non-optional group

baryluk:
Bug introduced in https://github.com/openZH/covid_19/pull/478
made some cantons loose time of a day information, and one of them
actually fail to parse due to the suffix.

Sorry.


baryluk:
I will add tests (both date and time parsing, and general field parsing / export format) for parser in a separate PR.

PR #487, opened 2020-04-09
Date parsing tests

baryluk:


metaodi:
Maybe we should put the tests in its own file and run the tests in the lint_python workflow.

Otherwise these tests will fail arbitrary scraper runs because of code changes.

baryluk:
> Maybe we should put the tests in its own file and run the tests in the lint_python workflow.

I don't know how to do that.

> Otherwise these tests will fail arbitrary scraper runs because of code changes.

Well, i felt that if somebody changes the code, and it fails, it is not going to be committed into repo? Exactly what tests are in my opinion. (Not to test the code, but to fail if incorrect code change is introduced).

PR #488, opened 2020-04-09
Extracting Hospitalized and ICU for ZG

dominikgehl:


scrapers/scrape_zg.py
```diff
@@ -1,21 +1,46 @@
...
+print('Date and time:', sc.find(r'Stand:? ([^<]+ Uhr)<', d))
 
-# 2020-03-23
-"""
+detailed_stats = soup.find("a", text=re.compile("Detaillierte Statistik"))
+if detailed_stats:
+    d_detailed_stats = sc.download(detailed_stats['href'])
+    d_detailed_stats_comps = urlparse(detailed_stats['href'])
+    csv_path = sc.find(r'csv_path:"([^"]+)"', d_detailed_stats)
+    csv_url = f"{d_detailed_stats_comps.scheme}://{d_detailed_stats_comps.netloc}{csv_path}"
```
baryluk, line 22: Just use `https://www.zg.ch`  There is no need for `urlparse` at all.

scrapers/scrape_zg.py
```diff
@@ -1,21 +1,46 @@
...
+    d_csv = sc.download(csv_url)
+    df = pd.read_csv(StringIO(d_csv), sep=",", parse_dates=['Datum'], dayfirst=True)
```
baryluk, line 24: Wouldn't using just `csv` (standard Python module) module be easier than pandas? I don't have pandas even installed.

dominikgehl: I found pandas just quicker for me to implement but give me a couple of minutes. I'll convert it to csv

baryluk: Well, I only now noticed that the data is transposed. So, it might be easier with pandas. Whatever looks cleaner in code.

cybermax: Pandas read_csv function is fantastic for reading and parsing csv files. It's a one-liner ;)

scrapers/scrape_zg.py
dominikgehl: Changed

scrapers/scrape_zg.py
dominikgehl: Converted it to CSV ... not much more complicated in this case

baryluk: Ok. Thanks.

baryluk:
LGTM.

PR #498, opened 2020-04-09
Various small changes in parser and expectation matrix code.

baryluk:
Use `=` not `startswith`, with only few exceptions.

This is to allow extended numbers parsing, like maybe 'Confirmed cases
(non-residents)' separately.

Fix output of parsing '2020-03-01' to be '2020-03-01T'.

Match a dot in the scrape timestamp, for parts of a second, in the CSV
converter.

Reduce redundancy when parsing extras in the CSV convert.

Use `extras_dict` directly to pass extras to `check_expectations`.

Check for expected time of a day.

Enable 'SH' to have time in the expectation matrix. This was because of
recent commit to fix parsing of the date/time format of SH.

Print warning to stderr if one of the values from the scraper is
literally string `'None'`.

### DIFF
--- a/scrapers/convert_parsed_to_csv.py
+++ b/scrapers/convert_parsed_to_csv.py
@@ -45,7 +45,7 @@ for line in sys.stdin:
   # GE 2020-03-27T         1924      23 OK 2020-03-28T18:57:34+01:00 # Extras: ncumul_hosp=313,ncumul_ICU=54 # URLs: https://www.ge.ch/document/point-coronavirus-maladie-covid-19/telecharger
 
   # Groups:             1              2                                         3       4              5                               6                            7             8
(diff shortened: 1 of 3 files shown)


baryluk:
Initial chunk of various changes. There is more coming. Including supporting v2 CSV format in the converter, failing on missing extras, and generally cleaning the handling of extras a bit further to be simpler and shorter (using dicts, instead of individual variables).

metaodi:
@baryluk did you notice the `old2newcsv.py` in the scripts directory? I thought you could just pipe the old output through this script and you're done for the csv part.

baryluk:
@metaodi Yes, they script works. Thanks. I guess, yes, just supporting new format will work then. Rest of stuff I mentioned stands. :)

metaodi:
Please rebase to fix the validation errors.

baryluk:
Rebased. I guess. I think. Looks good.

baryluk:
Thanks Stefan. Have a nice evening.

PR #517, opened 2020-04-12
Extracting data for Zug even if CSV file isn't up 2 date

dominikgehl:


metaodi:
Thanks for your quick response.

I can see a pattern here for all those cantons, that provide a file (e.g. GL or SZ), that on weekends only the website is updated, but not the file.
If this is reoccurring, I'd propose to skip the files and use again only the websites.

metaodi:
But I see that we actually lose information if we do that. Well yeah, let's try it like that and see how it works.

PR #520, opened 2020-04-13
Quick way of validating current parser output

dominikgehl:
Execute all parsers and compare output to expectation matrix

### DIFF
--- a/scrapers/scrape_matrix.py
+++ b/scrapers/scrape_matrix.py
@@ -12,108 +12,110 @@ import sys
 # Number of confirmed cases is always required to be present,
 # and is not listed in this list.
 matrix = {
-  # Note: Please keep the order of cantons and entries.
-  'AG': ['Deaths', 'Released', 'Hospitalized', 'ICU', 'Vent'],
-  'AI': [],
-  'AR': ['Deaths'],
-  'BE': ['Deaths', 'Hospitalized', 'ICU', 'Vent'],
-  'BL': ['Deaths', 'Released', 'Hospitalized', 'ICU'],
-  'BS': ['Deaths', 'Released', 'Hospitalized', 'ICU'],
-  'FR': ['Deaths', 'Released', 'Hospitalized', 'ICU'],
-  'GE': ['Deaths', 'Hospitalized', 'ICU'],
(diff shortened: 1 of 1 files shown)


metaodi:
I see how this is useful. I don't know yet where/when to use this script. If it's some kind of test, I'd use `assert` to fail if an expectation i  not met. Or at least fail with `sys.exit(1)` so that a caller knows it failed  (e.g. if we want to integrate it in the GitHub action workflow).

WDYT?

dominikgehl:
@metaodi Sure, we I can add the sys.exit(1) without issues. Actually I was wondering if it wouldn't make sense to switch to a more object orient design, creation an abstract class CantonParser and then have AGParser, VDParser, ZHParser etc that would inherit from them. We could then have the validation method in the parent class and only would need in each of the child classes to define the expected fields (similar to the matrix) ... but with the additional benefit of when running the parser we could immediately report that something is missing. Let me know if that's something you would be interested in.

dominikgehl:
For now I've updated with `sys.exit(1)` if any error is present. Let me know if you like the object orient approach and I'll start the refactoring

metaodi:
@dominikgehl I don't think this is a direction we currently want to go. While I see your point of less duplicate code and clearer structure, I think the simplicity of our current approach is a strength.

I think it's good, that a lot of people (even if they don't know python) are able to understand and contribute to the current code.

dominikgehl:
@metaodi Got it with regard to the refactoring. That's exactly why I asked before starting to work on it. Have a good evening !

baryluk:
@dominikgehl 

Could you move this code, to a separate file maybe?

I wanted to keep the matrix just a code with expectations.

I think the code is useful, but it should be separate and has few bugs.

Change

```python3
         result = subprocess.run(['python', f'scrape_{canton}.py'], stdout=subprocess.PIPE)
```

to

```python3
         result = subprocess.run([f'./scrape_{canton.lower()}.py'], stdout=subprocess.PIPE)
```

There is no need for `python`. In fact it is incorrect. As we use `python3`.

Also the cantons are in lower case letters in the file name.

Thank you.

dominikgehl:
@baryluk Sure. I'll move the code and create a new pull request

PR #526, opened 2020-04-14
Moving scraper validation to its own script

dominikgehl:
as suggested by @baryluk


baryluk:
Looks perfect. Thanks!

PR #535, opened 2020-04-15
Allow German words in BS parser for ICU number

dominikgehl:


scrapers/scrape_bs.py
```diff
@@ -112,7 +112,9 @@
   print('WARNING: Main pattern for matching hospitalized numbers failed to match', file=sys.stderr)
 
 print('Recovered:', sc.find(r'\b([0-9]+)\s*Personen\s*der\s*[0-9]+\s*positiv\s*Getesteten\s*.+\s*sind\s*wieder\s*genesen', d))
-print('ICU:', sc.find(r'Insgesamt\s*([0-9]+)\s*Personen benötigen\s*Intensivpflege', d))
+icu_number_exp = '([0-9]+|' + "|".join(sc.spelledOutNumbersMap.keys()) + ')'
```
metaodi, line 115: Is it really necessary to build this regex? Maybe you could just extract whatever is between "Insgesamt" and "Personen" and call sc.int_or_word on it. If it's not None, it was successful.

WDYT?

dominikgehl: Sure, that's also possible

dominikgehl:
@metaodi I've implemented the changes you suggested.

PR #541, opened 2020-04-17
Adjust xlsx parsing for GL. Extract time of day. Shift columns.

baryluk:
New column in GL spreadsheet with time of a day.


metaodi:
I just heard that SZ will add a time column as well. I think I will change the xls-parser to actually parse the column name as well (just like the DictReader from the csv module), so that we don't have to rely on the order of the columns.

WDYT?

baryluk:
@metaodi That would be awesome!

metaodi:
Here it is: #547

PR #547, opened 2020-04-17
Parse xls as dict

metaodi:
the parse_xls method aims to provide a similar functionality than the
DictReader of the built-in csv module. As long as we have a header row
and all the data below, we can build a list of dicts. This makes the
parsing prone to errors like new columns being added or the order of the
columns is changed.


metaodi:
@baryluk PTAL

metaodi:
I'm merging this since the SZ scraper currently fails.

PR #557, opened 2020-04-18
Add missing colon in ZG scraper

metaodi:


baryluk:
I just notice it too. Thank you for the fix.

PR #563, opened 2020-04-20
Adjusting to typo in BS webpage

dominikgehl:


scrapers/scrape_bs.py
```diff
@@ -114,6 +114,6 @@
 print(
     'Deaths:',
     sc.find(r'Basel-Stadt\s*verzeichnet\s*unverändert\s*([0-9]+)\s*Todesfälle', d) or
-    sc.find(r'Todesfälle\s*im\s*Kanton\s*Basel-Stadt\s*beträgt\s*nunmehr\s*insgesamt\s*([0-9]+)\b', d) or
+    sc.find(r'Todesfälle\s*im\s*Kanton\s*Basel-Stadt\s*beträgt\s*nu[nm]mehr\s*insgesamt\s*([0-9]+)\b', d) or
```
metaodi, line 117: Maybe we should remove the whole "nunmehr" part and just replace it with `.*`. WDYT?

dominikgehl: @metaodi Updated as you suggested. The regex now allows any (or no word)

PR #567, opened 2020-04-21
Extracting NE data from xlsx file

dominikgehl:


scrapers/scrape_ne.py
```diff
@@ -1,123 +1,18 @@
 #!/usr/bin/env python3
```
metaodi, line 1: Please add the following line after this:

```
# -*- coding: utf-8 -*-
```

dominikgehl: Done

PR #589, opened 2020-04-24
Extracting confirmed cases for NE from PDF

dominikgehl:


metaodi:
My understanding is, that it's on purpose that NE does not provide case numbers for the current day. Rather the numbers are updated later.

fyi: I'm currently working on a new kind of scraper to fetch whole time series to keep our data up-to-date with the published data from the cantons.

dominikgehl:
@metaodi Let me know if I can help with the new scraper

metaodi:
@dominikgehl yes absolutely, when I finish #600, I'll let you know. Then it's about extending all existing scrapers, that have access to a time series, to actually report the potential updated entries.

metaodi:
I've finished #600 and started to adapt the scrapers that have a time series, see the column "Time series" in the STATUS.md for details (https://github.com/openZH/covid_19/blob/master/scrapers/STATUS.md).

Feel free to adapt other scrapers as well.

The basic idea of the new setup is to use the existing scripts and simply produce more lines (one for each day) and then let the parse_scrape_output.py and add_to_db.py script handle every line.

I use 10 dashes (`----------`) as the delimiter between records. I already adapted the easy ones, that have an xls as source (GL, NW, NE, SH, JU, SZ, FR), use them as a template for further adaptions.

metaodi:
@dominikgehl do you agree that we can close this PR?

dominikgehl:
@metaodi Yes, agreed

PR #607, opened 2020-04-25
Extracting case numbers for previous days in AG

dominikgehl:
only case numbers are available for previous days

### DIFF
--- a/scrapers/scrape_ag.py
+++ b/scrapers/scrape_ag.py
@@ -3,7 +3,6 @@
 import scrape_common as sc
 import re
 
-print('AG')
 
 # get latest from list with all bulletins
 d = sc.download('https://www.ag.ch/de/themen_1/coronavirus_2/lagebulletins/lagebulletins_1.jsp')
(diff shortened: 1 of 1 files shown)


scrapers/scrape_ag.py
```diff
@@ -3,17 +3,31 @@
...
 # get latest from list with all bulletins
 d = sc.download('https://www.ag.ch/de/themen_1/coronavirus_2/lagebulletins/lagebulletins_1.jsp')
 
 url = sc.find(r'<a [^>]*href="([^"]+\.pdf)">.+Bulletin.+</a>', d)
 
 # download latest PDF
-d = sc.pdfdownload('https://www.ag.ch' + url, raw=True)
-
+pdf_url = 'https://www.ag.ch' + url
+d = sc.pdfdownload(pdf_url, raw=True)
```
metaodi, line 14: use silent=True to avoid output for this download.

dominikgehl: @metaodi Changed

PR #608, opened 2020-04-25
Extracting multiple days for Bern

dominikgehl:


scrapers/scrape_be.py
```diff
@@ -3,9 +3,8 @@
 from bs4 import BeautifulSoup
 import scrape_common as sc
 
-print('BE')
-d = sc.download('https://www.besondere-lage.sites.be.ch/besondere-lage_sites/de/index/corona/index.html')
-sc.timestamp()
+html_url = 'https://www.besondere-lage.sites.be.ch/besondere-lage_sites/de/index/corona/index.html'
+d = sc.download(html_url)
```
metaodi, line 7: use silent=True to avoid output from this download

dominikgehl: @metaodi Changed

PR #609, opened 2020-04-25
Extracting multiple days for VD

dominikgehl:


simgraworldwide:
Not sure where the scraper gets its data atm but I noticed that they do not update their table at https://www.vd.ch/toutes-les-actualites/hotline-et-informations-sur-le-coronavirus/point-de-situation-statistique-dans-le-canton-de-vaud/ regularly (despite the graphs being updated).

However, there is a xlsx table that can be accessed via https://partage.vd.ch/userportal/#/shared/public/hUApwTjAKdaXQCnB/StatistiquesDSAS

I recommend using this new source.

dominikgehl:
@simgraworldwide Sure, I can update the script for that xlsx file. But I'm surprised that the numbers are different in the file than in the webpage. Do we know why that's the case ?

simgraworldwide:
No. They seem to have a mess with their data, they change complete time series on a daily basis (I manually updated the data in the past couple of days and I do not have an explanation for this, it is really really weird, check the history of the file if you want to see what I mean), we are in contact with them, but no info so far.

Some thoughts why we should switch:
First, they did update their graphs (with data for 2020-04-23), which are, as far as I know, based on the xlsx I mentioned above. 
Further we have seen a lag in the update of their webpage compared to the xlsx sheet in the last 3 days (usually about 3-4 hours later).
So my guess is that they have just forgotten to update the table on their webpage, which would absolutely not surprise me, given the entire situation at VD (from my outsider perspective)...

dominikgehl:
Ok. Working on an update for the PR

dominikgehl:
Ah, that download page contains mostly JS ... it'll take me a bit more time to figure it out. Should be able to work on it this PM and then report back.

metaodi:
I converted the PR to "draft", click "Ready for review" when it's ready 😉

scrapers/scrape_vd.py
```diff
@@ -1,41 +1,77 @@
...
+    headers = rows[0]
+    assert headers == {'A': 'Date', 'B': 'Hospitalisation en cours', 'C': 'Dont soins intensifs', 'D': 'Décès', 'E': 'Nombre total de cas confirmés positifs'}
+    is_first = True
+    for row in rows[1:]:
+        if row['A'] is not None and isinstance(row['A'], datetime.datetime):
+            if is_first:
+                is_first = False
+            else:
+                print('-' * 10)
+            print('NE')
```
metaodi, line 66: This should be `VD`

dominikgehl: Good catch ! Fixed

scrapers/scrape_vd.py
```diff
@@ -1,41 +1,77 @@
...
+        print('Hospitalized:', cells[1])
+        print('ICU:', cells[2])
+        if cells[3].isnumeric():
+            print('Recovered:', cells[3])
+
+
+def parse_xlsx():
+    xls_url = 'https://partage.vd.ch/fss/public/link/public/stream/read/G12_HOP_POST_EPID_OMC_HOSP_SI_TA.xlsx?linkToken=hUApwTjAKdaXQCnB&itemName=StatistiquesDSAS'
+    xls = sc.xlsdownload(xls_url, silent=True)
+    rows = sc.parse_xls(xls)
```
metaodi, line 56: The header is on the third row, so passing `header_row=2` to `parse_xls`  leads to dicts like that:

```python
{'Date': datetime.datetime(2020, 4, 23, 0, 0), 'Hospitalisation en cours': 195, 'Dont soins intensifs': 44, 'Décès': 331, 'Nombre total de cas confirmés positifs': 5143}
```

Then you can use those names instead of A, B, C, D.

scrapers/scrape_vd.py
```diff
@@ -1,41 +1,77 @@
 #!/usr/bin/env python3
```
metaodi, line 1: Add `# -*- coding: utf-8 -*-` here

scrapers/scrape_vd.py
dominikgehl: Excellent. Changed

scrapers/scrape_vd.py
dominikgehl: Done

metaodi:
Great, thanks a lot!

PR #620, opened 2020-04-27
Bugfix: don't print record delimiter after last row

dominikgehl:


scrapers/scrape_sh.py
```diff
@@ -78,8 +82,3 @@
         print('Hospitalized:', (row['Hospitalisiert_Iso'] + row['Hospitalisiert_Intensiv']))
         print('ICU:', row['Hospitalisiert_Intensiv'])
     print('Deaths:', row['Verstorben'])
-    # do not print record delimiter for last record
-    # this is an indicator for the next script to check
-    # for expected values.
-    if len(rows) - 1 > i:
-        print('-' * 10)
```
metaodi, line 85: This part here takes care of that, or am I missing something?

PR #630, opened 2020-04-29
DayData class

dominikgehl:


scrapers/scrape_matrix.py
```diff
@@ -23,7 +23,7 @@
     'GR': ['Confirmed cases', 'Deaths', 'Hospitalized'],
     'JU': ['Confirmed cases', 'Deaths', 'Hospitalized', 'ICU'],
     'LU': ['Confirmed cases', 'Deaths', 'Hospitalized', 'ICU'],
-    'NE': ['Hospitalized', 'Vent'],
+    'NE': ['Confirmed cases', 'Deaths', 'Hospitalized', 'ICU', 'Vent'],
```
metaodi, line 26: This will fail most of the time, since NE does not publish this data every day. So if you define it as an "expectation" here, the scraper will fail and update no data.

dominikgehl: Reverted

metaodi:
Since this is so common to check if a value is `not None` and then print it, we could add a method to `scrape_common.py`, something like `printif(msg, value)`, to do this check. Then we don't have to throw around if statements all over the scrapers.
WDYT?

dominikgehl:
@metaodi The printif is definitely an option. But I would then rather go one step further and create a DayData class to hold the canton, source URL, timestamp and the different data points. The print method could then do all the necessary checks.

dominikgehl:
@metaodi Just added a commit with how this class could look like ...

metaodi:
Sounds good! I think the codebase is now mature enough to introduce some classes 😉

PR #682, opened 2020-05-07
Scraper SO date time fix

reberan:
replaced wrong character with :


metaodi:
The date/time scraping is all here: https://github.com/openZH/covid_19/blob/master/scrapers/scrape_dates.py

There are even some tests here: https://github.com/openZH/covid_19/blob/master/scrapers/test/test_dates.py

PR #687, opened 2020-05-08
Adding missing deaths column to AI expectation

dominikgehl:


metaodi:
Thanks, totally missed that.

PR #724, opened 2020-05-14
Fix regex ICU and Vent for AG

fabian:
Screenshot 2020-05-14 at 16 10 33


metaodi:
Thanks a lot! That was fast 😊

Issue #714, opened 2020-05-12
Missing data lead to strange results


emanuele-guidotti:
It seems that missing data lead to strange results.

For example, when filling the missing values with the previous non-missing value I obtain many cases in which the number of confirmed cases is greater than the number of tests.

You might want to have a look here were it is provided a log for these kind of issues:
https://covid19datahub.io/articles/iso/CHE.html
I hope this tool can help you to prioritize the data issues.

Best
-- Emanuele

tlorusso:
Dear Emanuele

Only a few Cantons report the number of tests on a regular basis. For some, the number of tests even has just been reported a single time. Filing the timeseries with the only value which has been reported so far is not an option in these cases. E.g. for BS, by filling the timeseries, the number of tests you include for the 12.5.2020 is actually still the value for the 3.2.2020. 

https://github.com/openZH/covid_19/blob/master/fallzahlen_kanton_total_csv_v2/COVID19_Fallzahlen_Kanton_BS_total.csv#L6

![grafik](https://user-images.githubusercontent.com/17902112/81777866-0bd92c80-94f2-11ea-9207-a316bac2f824.png)

emanuele-guidotti:
Thank you for your reply.

After removing the filling, there seem to be a few bugs in the data still.
We updated the logs with the raw data and I opened the issue #723 (no more related to missing data).

It would be very nice if you could list the project in the _Community Contributions_ of your README.
https://github.com/covid19datahub/COVID19/

Best,
-- Emanuele

metaodi:
@emanuele-guidotti please create a pull request for the README

PR #727, opened 2020-05-14
New community contribution

emanuele-guidotti:
Hello, 
It would be very nice if you could list our contribution. See #714 

Best,
-- Emanuele


metaodi:
Thank you!

PR #729, opened 2020-05-15
Adapting BL parser

dominikgehl:
website html changes

### DIFF
--- a/scrapers/scrape_bl.py
+++ b/scrapers/scrape_bl.py
@@ -90,7 +90,7 @@ Datum, Geheilte kalkuliert, Aktive Fälle, Todesfälle
 
     d = d.replace('\n', ' ')
     # Extract last line. Use non-greedy matching.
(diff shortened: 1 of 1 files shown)


scrapers/scrape_bl.py
```diff
@@ -125,6 +126,21 @@
                             rows[row_date]['cases'] = int(c[1] or 0) + int(c[2] or 0) + int(c[3] or 0)
                         rows[row_date]['recovered'] = c[1]
                         rows[row_date]['deaths'] = c[3]
+            else:
+                data = sc.find(r'<pre id="data_1".*?> ?Datum,&quot;Normale Station&quot;,&quot;Intensivstation \(nicht beatmet\)&quot;,&quot;Intensivstation \(beatmet\)&quot;\s*([^<]+)</pre>', d)
+                if data:
```
metaodi, line 131: instead of `if data` and the error message in the `else`, I'd use `assert data, "bug in regex"`

dominikgehl: good idea ! -> changed

PR #742, opened 2020-05-18
Fix scraper AG

fabian:
Do we have some clever solution for parsing hyphenation in other parsers?


fabian:
Closed in favour of https://github.com/openZH/covid_19/pull/743.

PR #749, opened 2020-05-19
Adapting AG scraper to PDF changes

dominikgehl:


metaodi:
Thank you @dominikgehl !

PR #750, opened 2020-05-19
Check that scraper is executable before trying validation

dominikgehl:
in order to skip GE for now


metaodi:
Sorry about that, that was my workaround to get https://github.com/openZH/covid_19/blob/master/scrapers/test_scraper.sh to work.

PR #751, opened 2020-05-19
Confirmed cases is always present for VS

dominikgehl:


metaodi:
When I look at the commit where I removed this 14 days ago (https://github.com/openZH/covid_19/commit/62e6ff0791334c4169fc3d91aa96433b8bf6488d) I'm pretty sure it was missing on that day. I'm fine with adding it again.

But in general I would focus on adapting the scrape_matrix for those cantons, that do not provide an XLS or CSV, since then we're protected from missing changes on the website. If already structured data is provided, it's obvious which fields are there and which are empty, so having one less expectation is fine.

PR #755, opened 2020-05-20
Remove confirmed cases from VS

metaodi:


metaodi:
@dominikgehl today they are missing, so I'm removing the expectation again.

PR #759, opened 2020-05-22
Converting SO scraper to DayData

dominikgehl:
so that we don't print the empty Hospitalized lines any more for recent days


metaodi:
The test failed, you need to skip the first line with "bis 31.03.2020"

dominikgehl:
@metaodi Sorry about that. I've added a fix.

Issue #767, opened 2020-05-26
SH stats reporting a decrease in "ncumul_conf"


nik-io:
Hi and thank you for the great work!

I have a personal script to plot the rate increase per day and got an assert failing that for Canton SH the reported "ncumul_conf" have decreased by 1 between 25/5 and 26/5. Not sure if this is an error in the source data but thought it'd make sense to report.

```
$ tail -n2 ../fallzahlen_kanton_total_csv_v2/COVID19_Fallzahlen_Kanton_SH_total.csv
2020-05-25,09:30,SH,,79,,5,1,,,6,https://sh.ch/CMS/get/file/5c0b7f8d-57b8-40ca-b145-7bf0a51d16a3
2020-05-26,09:30,SH,,78,,8,2,,,6,https://sh.ch/CMS/get/file/5c0b7f8d-57b8-40ca-b145-7bf0a51d16a3
```

Thanks again for this great work!

dominikgehl:
Yeah, I checked the original xlsx file and the line looks like

25.05.20 | 9.30 h | 79 | 4 | 1 | 6 | (1 Fall falsch positiv)

So there was apparently a false positive test ...

nik-io:
Thanks!

PR #768, opened 2020-05-26
SH - Subtract false positives

dominikgehl:
related to #767


metaodi:
I generally like the idea, but this should be fixed at the source. We're in contact with SH, so this "hack" should not be needed.

metaodi:
SH fixed their data

PR #786, opened 2020-06-02
Adding quarantined and isolated to AR scraper

dominikgehl:


metaodi:
@dominikgehl thank you for this! I'll run some tests with this, but looks good so far.

PR #789, opened 2020-06-02
Parsing isolated and quarantined for ZH

dominikgehl:


andreasamsler:
Thanks @dominikgehl for the parser.

@tlorusso and I decided to create a separate file for the respective ZH data: https://github.com/openZH/covid_19/blob/master/COVID19_Fallzahlen_Kanton_ZH_isolated_quarantined.csv

While our colleagues @GD-ZH are updating https://github.com/openZH/covid_19/blob/master/fallzahlen_kanton_total_csv_v2/COVID19_Fallzahlen_Kanton_ZH_total.csv daily, we will run a script, which is adding the `current_isolated` and `current_quarantined` to that file after each update.

PR #791, opened 2020-06-02
Adding Isolated and Quarantined to BS

dominikgehl:


andreasamsler:
BS is already doing this. Would the parser make your work easier, @jb3-2 ?

jb3-2:
Yes it would make my life easier! Ideally the prser would also calculate the other BS specific columns to the right of source.

scrapers/scrape_bs.py
```diff
@@ -125,4 +125,7 @@
     sc.find(r'Todesfälle\s*im\s*Kanton\s*Basel-Stadt\s*beträgt(?:\s*\S+)?\s*insgesamt\s*([0-9]+)\b', d) or \
     sc.find(r'Die\s*Zahl\s*der\s*Todesfälle\s*im\s*Kanton\s*Basel-Stadt\s*beträgt\s*.*unverändert\s*([0-9]+)\b', d)
 
+dd.isolated = sc.int_or_word(sc.find(r'\s+(\S+)\s+aktiven\s+Fällen\s+in\s+Isolation', d))
```
metaodi, line 128: @dominikgehl the text today is a bit different

> Die Zahl der 979 Infektionen setzt sich zusammen aus 923 genesenen Personen (-), 50 Todesfällen (-) und sechs aktiven Fällen (+ 1).

So for the `current_isolated` the part "in Isolation" is optional

dominikgehl: updated

jb3-2:
Thank you very much!!

metaodi:
@jb3-2 maybe we can get those other numbers as well. Currently the scraper scripts are not equipped to handle variables that only occur for one canton. Could you maybe create an issue for that?

jb3-2:
@metaodi OK I have created an issue: https://github.com/openZH/covid_19/issues/796 - thanks! J.

PR #803, opened 2020-06-05
Correctly indicate data source URL for VS

dominikgehl:


metaodi:
This is on purpose, since VS provides information on their website, but delivers their XLS by email which then gets uploaded to GitHub. VS asked us to provide their website as source.

Issue #797, opened 2020-06-04
AR scraper distinguishes between case and contact tracing numbers


andreasamsler:
Timestamps of data about cases and about contact tracing communicated by AR on https://www.ar.ch/verwaltung/departement-gesundheit-und-soziales/amt-fuer-gesundheit/informationsseite-coronavirus/ does vary.

Currently the scraper is adding a new record with both case numbers and contact tracing numbers with the latest timestamp on the webpage.

Preferably the scraper is adapted, so that it is sensitive with regard to timestamps of the two distinct groups of variables.

metaodi:
@andreasamsler do you take care to fix the "wrong" data in the CSV of AR? I guess there should be no case data for 2020-06-03 and 2020-06-04, but I'm not entirely sure.

PR #805, opened 2020-06-05
Adding Isolated and Quarantined for AR

dominikgehl:


metaodi:
This was separted in two rows, see #797 and #798

PR #834, opened 2020-06-15
Minor change in quarantine data from Zug

statzg:
We had to remove the distinction between quarantined in home and out of home, since the data ist not really available. So the scaper gets simplified.


metaodi:
Oh sorry didn't see that before, so I merged #835 instead. Thanks for PR anyway!

PR #856, opened 2020-06-23
AG does not provide isolated/quaratined every day

metaodi:


fabian:
@metaodi I think they actually provide it each day, I think the scraper should be fixed instead.

metaodi:
oh sorry, I thought they have different timestamps, but apparently this is not the case.

fabian:
Yeah the table/layout is horrible… will send a PR with fix.

Issue #863, opened 2020-06-24
Geneva PDF file contains newer data than parsed xlsx


dominikgehl:
https://www.ge.ch/document/covid-19-situation-epidemiologique-geneve links to both a PDF and an xlsx file. The PDF file seems to contain more up to date information. Do we want to parse both ?

dominikgehl:
@andreasamsler Did you get any news from Geneva ? Should we start trying to parse the data from the PDF ?

dominikgehl:
Latest PDF version is now https://www.ge.ch/document/20094/telecharger and contains data from today

andreasamsler:
GE updated the XLS today; 'ncumul_deceased' is empty. We notified GE, and fix the missing values for the moment.

PR #867, opened 2020-06-25
parse latest weekly pdf and get number of tests for GE

maekke:
Related to #863, but parses another document (since the mentioned doesn't really contain any text, but just images).
Not 100% sure if things are fine as is, but it should parse the latest pdf and extract the number of tests


andreasamsler:
AFAWK, the BAG is going to publish updated numbers of tests p. all Cantons. We will check by the beginning of next week. Best, Andreas

maekke:
> AFAWK, the BAG is going to publish updated numbers of tests p. all Cantons. We will check by the beginning of next week. Best, Andreas

Sure, that would be nice as well!

PR #870, opened 2020-06-27
Skip row for BE

metaodi:


dominikgehl:
I took your PR as a starting point and added another small adjustment in #874

Issue #875, opened 2020-06-29
New datasource for BS


dominikgehl:
https://data.bs.ch/explore/dataset/100073/table/?sort=timestamp

as mentioned here https://www.gd.bs.ch/nm/2020-coronavirus-unveraenderte-fallzahlen-im-kanton-basel-stadt--tagesbulletin-wird-zum-wochenbulletin-gd.html

metaodi:
See https://github.com/openZH/covid_19/pull/876#issuecomment-652252619

PR #876, opened 2020-06-29
BS new datasource including full history

dominikgehl:
for #875

### DIFF
--- a/scrapers/scrape_bs.py
+++ b/scrapers/scrape_bs.py
@@ -2,150 +2,43 @@
 # -*- coding: utf-8 -*-
 
 import re
-import sys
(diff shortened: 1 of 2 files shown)


metaodi:
@dominikgehl I just asked @jb3-2 from BS wheter their data portal is indeed a "new datasource". It turns out, that they manually update the CSV in this repo here and then copy it over to their data portal. So we can't really scrape this data since the source is currently this very repo. So unfortunately I have to close this PR 😞 . According to them, it's more likely that we can parse press releases again in the future, should they decide to publish them more often.

jb3-2:
@dominikgehl Thanks for your work anyway! Greatly appreciated.

PR #878, opened 2020-06-29
Removing some warning from the validation script

dominikgehl:


metaodi:
I propose to remove the warning for those "unexpected inputs". Since this is not really an error, but rather means that a canton does not always provide a number, but when they do it's fine.

dominikgehl:
> I propose to remove the warning for those "unexpected inputs". Since this is not really an error, but rather means that a canton does not always provide a number, but when they do it's fine.

Updated

PR #889, opened 2020-07-06
Adding Isolated and Quarantined for LU

dominikgehl:
#888

### DIFF
--- a/scrapers/scrape_lu.py
+++ b/scrapers/scrape_lu.py
@@ -76,18 +76,27 @@ dd = sc.DayData(canton='LU', url=url)
 
 include_hosp = True
 include_cases = True
+include_isolated = True
 
 case_date_str = sc.find(r'Fallzahlen\s*im\s*Kanton\s*Luzern.*\(Stand:\s*(.+?)\,', d)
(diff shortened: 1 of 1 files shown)


scrapers/scrape_lu.py
```diff
@@ -76,18 +76,27 @@
...
 case_date_str = sc.find(r'Fallzahlen\s*im\s*Kanton\s*Luzern.*\(Stand:\s*(.+?)\,', d)
 hosp_date_str = sc.find(r'Hospitalisierungen.*\(Stand:\s*(.+?)\,', d)
+isolated_date_str = sc.find(r'Isolation.*\(Stand:\s*(.+?)\,', d)
 
 case_date = sc.date_from_text(case_date_str)
 hosp_date = sc.date_from_text(hosp_date_str)
-if case_date > hosp_date:
+isolated_date = sc.date_from_text(isolated_date_str)
+
+max_date = max(isolated_date, case_date, isolated_date)
```
metaodi, line 89: one of the two `isolated_date`s should be `hosp_date` I think.

dominikgehl: good catch. Should be fixed now

PR #908, opened 2020-07-15
add a scraper for performed tests

maekke:
Kind of a "complete" follow up of #867 but for all cantons / data from BAG.
It should basically work, but I'm not sure if this should be moved into the common module and be wired up in the individual canton scrapers?
Right now, it's not yet wired up anywhere, so it's not yet very useful...
Inputs are welcome!

### DIFF
--- a/scrapers/scrape_tests.py
+++ b/scrapers/scrape_tests.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python3
+
+import scrape_common as sc
+import sys
+import re
+
+
+# download latest PDF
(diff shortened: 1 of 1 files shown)


metaodi:
Happy to merge this, but I'm not sure how/if we wire this with the CSVs of this repo. We decided a while ago to not mix cantonal data with BAG data.

But it's great to have the script, so others can make use of it. And it's very easy to integrate, should we decide to later on. Maybe this could be a separate CSV?

PR #910, opened 2020-07-16
add number of tests for ZH scraper

maekke:


scrapers/scrape_zh.py
```diff
@@ -43,6 +43,15 @@
 
 print(dd_iso_q)
 
+dd_tested = sc.DayData(canton='ZH', url=url)
+date_time_info = sc.find(r'liessen sich vom \d+\. bis (\d+\. .* 20\d{2}) testen,', d)
+dd_tested.datetime = date_time_info
+dd_tested.tested = soup.find(string=re.compile(r'liessen sich vom')).find_previous('h4').text.replace('\'', '')
```
metaodi, line 49: This value does not represent cumulative tests until this date, but rather number of tests in this period. For this to work, you would need to sum this number over time. I'm not sure there is enough data available to do this.

maekke: You are right, I didn't think about that... I guess we can close this PR, since the data doesn't seem to be available.

PR #920, opened 2020-07-20
Parsing OW XLSX file

### LINKED ISSUE #916: New source for OW
OW publishes time series data as xlsx file, accessible via https://www.ow.ch/de/kanton/publired/publikationen/?action=info&pubid=20318. Scraper needs to be updated and the data adjusted accordingly.

dominikgehl:
Fixes #916

### DIFF
--- a/scrapers/scrape_ow.py
+++ b/scrapers/scrape_ow.py
@@ -1,66 +1,42 @@
 #!/usr/bin/env python3
 
 import re
+import datetime
 from bs4 import BeautifulSoup
 import scrape_common as sc
 
-print('OW')
-d = sc.download('https://www.ow.ch/de/verwaltung/dienstleistungen/?dienst_id=5962', encoding='windows-1252')
-sc.timestamp()
-d = d.replace('&nbsp;', ' ')
-
-# 2020-03-23
-"""
-<p class="object-pages-img"><img src="../../images/5e73948a8f49f.jpg"  alt="Kampagne BAG" style="width:600;height:293;border:0;" /></p><br /><div class="object-pages-description"><p class="icmsPContent icms-wysiwyg-first"><em>Stand 23.03.2020</em></p>
-...
-...
-<h3 class="icmsH3Content"><strong>Anzahl Infizierte</strong></h3>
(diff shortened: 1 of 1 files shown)


scrapers/scrape_ow.py
```diff
@@ -1,66 +1,42 @@
...
-print('Date and time:',
-      sc.find(r'Stand ([^<]+ Uhr)', d) or sc.find(r'Stand ([^<]+)<', d))
-
+d = sc.download('https://www.ow.ch/de/kanton/publired/publikationen/?action=info&pubid=20318',
+                encoding='windows-1252', silent=True)
 soup = BeautifulSoup(d, 'html.parser')
-i = 0
-for row in soup.find(id='Fallzahl').find_next('table').find_all('tr'):
-    i += 1
+xls_url = soup.find(href=re.compile("\.XLSX$")).get('href')
```
metaodi, line 11: Maybe make this case insensitive, just in case someone uploads a .xlsx file in the future

scrapers/scrape_ow.py
```diff
@@ -1,66 +1,42 @@
...
-    if i == 1: # Header.
-        assert 'COVID-19' in str(cells[0].find('strong').string)
-        assert 'Anzahl' in str(cells[1].find('strong').string)
-        # assert 'Veränderung' in str(cells[2].find('strong').string)
-        continue
+    if cells[0].string and cells[0].string.startswith('Datum'):
+        file_date = cells[1].string
+
+xls = sc.xlsdownload(xls_url, silent=True)
+rows = sc.parse_xls(xls, header_row=0)
```
metaodi, line 20: if you pass header_row=4, you should get nice headers (something like `row['Infizierte Personen']` instead of `row['B']`

scrapers/scrape_ow.py
dominikgehl: Changed

scrapers/scrape_ow.py
dominikgehl: Thanks for the suggestion. Updated

PR #946, opened 2020-07-31
ZG: supporting NA in 'Anzahl' column of csv file

dominikgehl:


scrapers/scrape_zg.py
```diff
@@ -78,18 +78,23 @@
...
     print('-' * 10)
     dd = sc.DayData(canton='ZG', url=main_url)
     dd.datetime = day
-    dd.cases = data[day]['Fallzahl']
-    dd.hospitalized = data[day]['Hospitalisierte']
-    dd.icu = data[day]['Hospitalisierte in Intensivpflege']
-    dd.recovered = data[day]['Genesene']
-    dd.deaths = data[day]['Todesfälle']
+    if 'Fallzahl' in data[day]:
+        dd.cases = data[day]['Fallzahl']
```
metaodi, line 90: instead of checking this for every field, simply use `dd.cases = data[day].get('Fallzahl')`

if `Fallzahl` i  not there, `None` will be returned.

dominikgehl: updated

metaodi:
Thanks  @dominikgehl

PR #953, opened 2020-08-03
add quarantined and isolated to VD scraper

maekke:
could be that the text in the PDF will change, so this might not be bullet proof...


maekke:
for some reason, the SwissSign certificate is not known - same on my machine... don't really know how to handle this best.

metaodi:
It seems that the SwissSign cert is missing in the CA-Bundle of Mozilla. I'll take care of it in #955

Issue #967, opened 2020-08-06
FL: PDF URL not found


dominikgehl:
Latest data I could find was linked from https://hebensorg.li/
Do we want to parse this source ? Do we have a better source of data ?

metaodi:
But https://www.llv.li/files/ag/aktuelle-fallzahlen.pdf seems to have current data. Maybe it was updated just now (?)

dominikgehl:
Scraper still returns empty though. Looking into it.

dominikgehl:
The PDF now contains two additional columns (recovered and deaths). scraper is updated.

PR #968, opened 2020-08-06
FL: adding recovered and deaths columns

dominikgehl:
#967

### DIFF
--- a/scrapers/scrape_fl.py
+++ b/scrapers/scrape_fl.py
@@ -35,11 +35,13 @@ else:
 history_url = 'https://www.llv.li/files/ag/aktuelle-fallzahlen.pdf'
 d = sc.pdfdownload(history_url, layout=True, silent=True)
 for row in d.splitlines():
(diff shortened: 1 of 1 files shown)


scrapers/scrape_fl.py
```diff
@@ -35,11 +35,13 @@
...
-    m = re.search(r'^(?:Montag|Dienstag|Mittwoch|Donnerstag|Freitag|Samstag|Sonntag),\s+(.+\d{4})\s+(\d+)$', row)
+    m = re.search(r'^(?:Montag|Dienstag|Mittwoch|Donnerstag|Freitag|Samstag|Sonntag),\s+(.+\d{4})\s+(\d+)\s+(\d+)\s+(\d+)$', row)
     if m:
         dd_full_list = sc.DayData(canton='FL', url=history_url)
         dd_full_list.datetime = m[1]
         dd_full_list.cases = m[2]
+        dd_full_list.recovered = m[3]
+        dd_full_list.deaths = m[4]
         print('-' * 10)
         print(dd_full_list)
```
metaodi, line 46: Maybe there should be an check after the loop to indicate an error if there was no match in any row, so we get notified earlier of a changed PDF. Probably some error message like above and `sys.exit(1)` should be enough

dominikgehl: Asserting added

PR #994, opened 2020-08-19
SH: fix parsing deaths

dominikgehl:


metaodi:
we're currently trying to convince SH to not add such comments directly in the number column, so that hopefully this "hack" is not needed.

dominikgehl:
> we're currently trying to convince SH to not add such comments directly in the number column, so that hopefully this "hack" is not needed.

Let's hope it won't be needed.

metaodi:
I'll merge this now, maybe I'll remove it later on, when the source is fixed.

PR #1008, opened 2020-08-21
add isolated and quarantined cases for SG

maekke:
I'm not 100% sure if the data is actually correct - the excel sheet on the page seems to have a bit different numbers for isolated and quarantined cases...


maekke:
The numbers from today seem to match regarding html and the excel file.

metaodi:
Thanks @maekke 

it seems that the xls is a bit more up to date regarding this data, so maybe we will switch to the xls eventually. But currently the website has IMHO the better structure than the xls, so let's leave it like that for now.

PR #1024, opened 2020-08-30

maekke:
In principle the older weekly reports could be scraped as well to get the  isolated/quarantined values from previous weeks as well... 
I guess this would be a bit an overkill to do this every 20mins. Maybe a daily job for things like that should be introduced? :thinking:


maekke:
and should I add the three values to the corresponding `DayData` entry (so scrape it before and match it while parsing the xls) in order to avoid the validation error?

maekke:
See #1025

PR #1025, opened 2020-08-30

maekke:
See #1024 - this "version" should address the failing test run.

As stated in the other PR, there would be more data available in the previous weekly reports. This could be scraped as well, but I have the feeling that it would be a bit overkill to do this every 20mins - maybe a daily job could be introduced for it? :thinking:

### DIFF
--- a/scrapers/scrape_vs.py
+++ b/scrapers/scrape_vs.py
@@ -13,7 +13,8 @@ content = sc.download(stat_url, silent=True)
 soup = BeautifulSoup(content, 'html.parser')
 res = soup.find(string=re.compile(r'Synthese COVID19 VS Woche\d+')).find_previous('a')
 weekly_pdf_url = base_url + res.attrs['href']
-content = sc.pdfdownload(weekly_pdf_url.replace(' ', '%20'), silent=True)
+weekly_pdf_url = weekly_pdf_url.replace(' ', '%20')
+content = sc.pdfdownload(weekly_pdf_url, silent=True)
 
 
 # add isolated/quarantined to the existing DayData item
--- a/scrapers/scrape_vs.py
+++ b/scrapers/scrape_vs.py
@@ -2,13 +2,15 @@
 # -*- coding: utf-8 -*-
 
 import datetime
+import re
+from bs4 import BeautifulSoup
 import scrape_common as sc
 
 xls_url = 'https://raw.githubusercontent.com/statistikZH/covid19_drop/master/Chiffres%20%20COVID-19%20Valais.xlsx'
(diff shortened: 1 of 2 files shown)


scrapers/scrape_vs.py
```diff
@@ -31,5 +29,30 @@
...
+soup = BeautifulSoup(content, 'html.parser')
+res = soup.find(string=re.compile(r'Synthese COVID19 VS Woche\d+')).find_previous('a')
+weekly_pdf_url = base_url + res.attrs['href']
+content = sc.pdfdownload(weekly_pdf_url.replace(' ', '%20'), silent=True)
 
+# add isolated/quarantined to the existing DayData item
+week_end_date = sc.find(r'vom (\d+)\. bis (\d+\.\d+\.20\d{2})', content, group=2)
+week_end_date = sc.date_from_text(week_end_date).isoformat()
+for dd in data:
+    if dd.datetime == week_end_date:
```
metaodi, line 47: Why do you loop over the `data` list? I would simply create new DayData objects with the week_end_date as it's `datetime`. There is no need to reuse the existing objects. Should for whatever reason a date be missing in the xls, this would fail to add the isolated and quarantined numbers.

maekke: I implemented it like that in #1024, but there was a problem with the CI check... I'm not really sure, what's wrong with the changes there (or the check is "wrong").

metaodi: Ah yes, the check is a bit stupid. It only runs on the last entry (as the assumption is, that the last entry is the most complete). This is mostly to be able to add an expected value (e.g. `deaths`) but not fail, if at any point in the past, this value was not provided.
I would simply change the order, so first scrape the weekly pdf, and then scrape the xls. That way, there shouldn't be a problem.

maekke: Ah that was it... I've created a new PR #1027

metaodi: why a new PR? you could just update this one. Then the conversation is not lost.

maekke:
Reopening...

scrapers/scrape_vs.py
maekke: To have a cleaner history. I've reopened it and moved the isolated/quarantined parsing to the top of the file.

metaodi: If you want, you can always rebase your branch, and remove squash etc. your commits. It's up to you. If you prefer to create a new PR, I don't mind. 👍

metaodi:
Thank you very much @maekke !

PR #1027, opened 2020-09-01

maekke:
next try, see #1025 .

### DIFF
--- a/scrapers/scrape_vs.py
+++ b/scrapers/scrape_vs.py
@@ -2,13 +2,31 @@
 # -*- coding: utf-8 -*-
 
 import datetime
+import re
+from bs4 import BeautifulSoup
 import scrape_common as sc
+
+# parse weekly data for isolated and quarantined numbers
+base_url = 'https://www.vs.ch'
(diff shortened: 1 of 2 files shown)


scrapers/scrape_vs.py
```diff
@@ -2,13 +2,31 @@
...
+from bs4 import BeautifulSoup
 import scrape_common as sc
 
+# parse weekly data for isolated and quarantined numbers
+base_url = 'https://www.vs.ch'
+stat_url = base_url + '/de/web/coronavirus/statistiques'
+content = sc.download(stat_url, silent=True)
+soup = BeautifulSoup(content, 'html.parser')
+res = soup.find(string=re.compile(r'Synthese COVID19 VS Woche\d+')).find_previous('a')
+weekly_pdf_url = base_url + res.attrs['href']
```
metaodi, line 15: Please replace the spaces here, so that the correct URL is used on line 18:

`(base_url + res.attrs['href']).replace(' ', '%20')`

metaodi:
so let's use #1025

PR #1033, opened 2020-09-05
GR: add scraper to for district data

maekke:
see #831 

I'm not sure yet how to integrate it to be scraped automatically...
The data format should be pretty similar to the Zürich Bezirke one.


metaodi:
I'll take care to integrate it next week.

maekke:
I've just pushed a small change to move the `DistrictData` class to scrape_common, so this can be used in other scrapers...
(For example BE and SG are providing district data as well)

PR #1043, opened 2020-09-09
Add FR district scraper

maekke:
since FR is only providing the last three weeks, add "incremental" data support to `run_district_scraper.sh ` (not perfect, but it should work). Also note that this will change the order of the file contents between the first and every upcoming run, which shouldn't be a big issue I suppose.


maekke:
Please wait with merging. The appending of data doesn't work, when data is changed (date/week identical, but for example deceased changes).

maekke:
7d8d082 should fix this issue (hopefully :sweat_smile: ).

metaodi:
@maekke maybe we should use the same mechanism for appending and updating data like the other scraper (using sqlite as an intermediate file format). WDYT?

maekke:
@metaodi this would probably be way better and more robust...

metaodi:
okay, I'll work on that.

PR #1076, opened 2020-09-21
Add VS district scraper

maekke:
I'm not 100% happy and confident with this one. The PDF parsing seems a bit fragile, but I didn't really find a better solution...

### DIFF
--- a/scrapers/scrape_vs_districts.py
+++ b/scrapers/scrape_vs_districts.py
@@ -1,0 +1,121 @@
+#!/usr/bin/env python
+
+import re
+import sys
+
+from bs4 import BeautifulSoup
+
+import scrape_common as sc
+
+# get the latest weekly PDF
+base_url = 'https://www.vs.ch'
+url = base_url + '/de/web/coronavirus/statistiques'
+content = sc.download(url, silent=True)
+soup = BeautifulSoup(content, 'html.parser')
+link = soup.find(href=re.compile(r'Synthese.*Woche'))
+
+# fetch the PDF
+url = base_url + link['href'].replace(' ', '%20')
+content = sc.pdfdownload(url, silent=True, page=1)
+week = sc.find(r'Epidemiologische Situation Woche (\d+)', content)
+year = sc.find(r'\d+\.\d+\.(\d{4})', content)
+
+content = sc.pdfdownload(url, silent=True, page=11, layout=True, rect=[0, 403, 450, 50])
+
+# kill the left most values
+content = re.sub(r'^\d+\s', ' ', content)
+content = re.sub(r'\n\d+ ', '\n ', content)
+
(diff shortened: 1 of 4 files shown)


metaodi:
It seems this does not yet work, e.g. the last PDF has 12 pages. Maybe always use the last page?

scrapers/scrape_vs_districts.py
```diff
@@ -0,0 +1,121 @@
...
+    47750,
+    28910,
+    47980,
+    15260,
+    13830,
+    46840,
+]
+
+
+if len(cases_sorted) == 13:
```
metaodi, line 107: Please use an assertion instead of if/else here, so that the scraper fails if there are not 13 values

```python
assert len(cases_sorted) == 13, f'expected 13 district values, but got {len(cases_sorted)} for {url}'
```

scrapers/scrape_vs_districts.py
```diff
@@ -0,0 +1,121 @@
...
+    # for i in range(13):
+    i = 0
+    for key, value in cases_sorted.items():
```
metaodi, line 110: if you use `for i, (key, value) in enumerate(cases_sorted.items())` you get the index variable `i` and the key/value pair, no need to count yourself.

metaodi (changes requested):

scrapers/scrape_vs_districts.py
maekke: Sure, I'll change it to an `assert`. But the main issue here remains, that for example this weeks bulletin will not provide 13 values, because the zeros are not in the PDF (they were last week).
I haven't yet found an automated way to process this properly... Do you have any idea @metaodi ?

maekke: Okay, the assert is now in place with the "functionality" to make it work with the 0 values not available... It should be more or less robust I hope :sweat_smile:

scrapers/scrape_vs_districts.py
maekke: I guess this is obsolete now.

maekke:
> It seems this does not yet work, e.g. the last PDF has 12 pages. Maybe always use the last page?

I guess this could be done, but then the `pdfdownload` functionality would need to be split up. Something like that:
- `pdf_file_download()` that returns the /tmp/xxx file name
- `get_pdf_pages()` that could use `pdfinfo` command
- `get_pdf_content()` with the current `pdfdownload` arguments.
- `pdfdownload` could be changed to use `pdf_file_download` and `get_pdf_content`
This could also be done in a separate PR...

maekke:
> It seems this does not yet work, e.g. the last PDF has 12 pages. Maybe always use the last page?

@metaodi this should now be in place...

PR #1081, opened 2020-09-27
use sqlite for district data combining

maekke:
it's kind of based on the logic of the normal scrapers, but a bit simplified (due to a "single source" of the data).
Also there were a few adjustments to the GR scraper, basically to drop the non-GR data entries.

Once this is one is reviewed/merged there might be small adjustments needed for the district scraper pull requests.

### DIFF
--- a/scrapers/add_district_db_entry.py
+++ b/scrapers/add_district_db_entry.py
@@ -1,0 +1,101 @@
+#!/usr/bin/env python3
+
+import sys
+import sqlite3
+import traceback
+import os
+
+import scrape_common as sc
+
+__location__ = os.path.realpath(
+    os.path.join(
+        os.getcwd(),
+        os.path.dirname(__file__)
+    )
+)
+
+input_failures = 0
+
+try:
+    DATABASE_NAME = os.path.join(__location__, 'data.sqlite')
+    conn = sqlite3.connect(DATABASE_NAME)
+
+    i = 0
+    for line in sys.stdin:
+        dd = sc.DistrictData()
+        if dd.parse(line.strip()):
+            c = conn.cursor()
+            try:
+                print(dd)
+
+                c.execute(
+                    '''
+                    INSERT INTO data (
+                      DistrictId,
+                      District,
+                      Canton,
+                      Date,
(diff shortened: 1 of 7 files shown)


metaodi:
This looks good, I hope I can test this this week, so we can finally start adding all those districts scrapers (thanks you!).

One thing: please remove the `fallzahlen_bezirke/fallzahlen_kanton_GR_bezirk.csv` from this PR, since it will most likely cause conflicts until it's merged.

maekke:
> This looks good, I hope I can test this this week, so we can finally start adding all those districts scrapers (thanks you!).
> 
> One thing: please remove the `fallzahlen_bezirke/fallzahlen_kanton_GR_bezirk.csv` from this PR, since it will most likely cause conflicts until it's merged.

Thanks for the review! I've dropped the GR bezirk csv commit, but that will be something you need to do manually after merging. Since `BezirkId` is `NOT NULL` (due to the `UNIQUE` condition) and will cause problems otherwise.
To avoid the manual editing, I could cherry pick 8f458bc to a separate PR...

scrapers/add_district_db_entry.py
```diff
@@ -0,0 +1,101 @@
...
+                      TotalConfCases,
+                      NewConfCases,
+                      TotalDeaths,
+                      NewDeaths,
+                      SourceUrl
+                    )
+                    VALUES
+                    (?,?,?,?,?,?,?,?,?,?,?,?)
+                    ON CONFLICT (DistrictId, District, Canton, Date, Week, Year)
+                    DO UPDATE SET
```
metaodi, line 50: Unfortunetely this does only work in newer versions of sqlite. I tried to use this UPSERT logic for the other scraper as well but had to give up, since it's a hassle to get the latest sqlite version on GitHub Actions.

Instead I opted for a version that works with all versions of sqlite by trying to INSERT, check if the unique constraints was violated (`except sqlite3.IntegrityError:`), and if so, use an UPDATE statement instead. It's not beautiful, but it works.

I will merge this PR and create a follow-up with the new INSERT/UPDATE logic.

PR #1098, opened 2020-10-05
JU: checking that URL can be found

dominikgehl:


metaodi:
fyi: I disabled JU for the moment.

PR #1106, opened 2020-10-05
use the date instead of the week number for SG district scraper

maekke:
The data is pretty unreasonable right now, which is due to the last week entry is used. Instead it's probably better to use the dates...
I guess all existing data should be removed after merging this one, to avoid week/date data mixup.

sg_districts


metaodi:
So do the district files have _either_ a date _or_ a week number, but never both?

maekke:
> So do the district files have _either_ a date _or_ a week number, but never both?

You're probably right, there is no real reason to not have both. I initially thought that it's kind of redundant, but I guess it doesn't harm to have both.
I've restored the week and added the year in cc0157b.

PR #1146, opened 2020-10-10
BE district scraper: improve city name matching logic

maekke:
- replace the known abbreviations to the proper name
- and keep only a subset of hard-coded city names that do not match the list

### DIFF
--- a/scrapers/scrape_be_districts.py
+++ b/scrapers/scrape_be_districts.py
@@ -26,6 +26,7 @@ def parse_kleinstgemeinde(item, dds):
     # handle kleinstgemeinde stuff
     district = sc.find(r'Kleinst.* im( Verwaltungskreis)? (.*)', item, group=2)
     if district is not None:
+        district = district.strip()
         if district == 'Berner Jura':
             district = 'Jura bernois'
         dds[district].new_cases += new_cases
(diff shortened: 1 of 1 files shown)


scrapers/scrape_be_districts.py
```diff
@@ -6,20 +6,30 @@
...
+    # and handle a few special cases
     cities = {
-        'Biel': 'Biel / Bienne',
-        'Wohlen b. B.': 'Wohlen bei Bern',
-        'Muri-Gümligen': 'Muri bei Bern',
         'St-Imier': 'Saint-Imier',
-        'Büren a.A.': 'Büren an der Aare',
-        'Langnau i.E.': 'Langnau im Emmental',
-        'Oberhofen': 'Oberhofen am Thunersee',
-        'Erlenbach': 'Erlenbach im Simmental',
```
metaodi, line 17: What about Biel, Oberhofen and Erlenbach? I think
 those should be kept.

metaodi: And Muri-Gümligen, too

maekke: No, this shouldn't be required anymore, unless for example multiple 'Erlenbach' would exist in the canton.
Sorry I forgot to mention this in the original comment, but on line 118+, this should now be handled by looking for the city name provided by the page in the available cities.
I'll re-add Muri-Gümligen, I removed this one by mistake...

maekke:
And added two more commits to fix today's data.

PR #1147, opened 2020-10-10
Scrape weekly GL pdf

maekke:
to get isolated, quarantined, travel quarantined and ICU values


fallzahlen_kanton_total_csv_v2/COVID19_Fallzahlen_Kanton_GL_total.csv
```diff
@@ -1,4 +1,4 @@
-date,time,abbreviation_canton_and_fl,ncumul_tested,ncumul_conf,new_hosp,current_hosp,current_icu,current_vent,ncumul_released,ncumul_deceased,source,current_isolated,current_quarantined
+date,time,abbreviation_canton_and_fl,ncumul_tested,ncumul_conf,new_hosp,current_hosp,current_icu,current_vent,ncumul_released,ncumul_deceased,source,current_isolated,current_quarantined,current_quarantined_riskareatravel
```
metaodi, line 1: You need to add the two new column on every row. There is a script to help with that: [`scripts/add_new_columns.py`](https://github.com/openZH/covid_19/blob/master/scripts/add_new_columns.py)

maekke: I see, thanks for the feedback! Should be fixed in 80e2aeb

PR #1177, opened 2020-10-14
add missing population value to BL district scraper

maekke:
not my day apparently :sweat_smile:


metaodi:
@maekke all good, thanks for taking care of all those issues. I didn't have the time to properly test your new scraper, but it's no problem to fix it later on.

PR #1183, opened 2020-10-15
BE district scraper: Avoid output with no data available

### LINKED ISSUE #1180: BE district scraper gives all "0" values
Homepage has this comment: "Aufgrund der hohen Anzahl Fälle werden keine Ortschaften genannt."

maekke:
is it fine like that? or would it make more sense to output the entries, but with empty `NewConfCases` value?
I'm not fully sure, what is more reasonable...

resolves #1180


metaodi:
I think this is the way to go, just like the other scraper.

dr-itz:
Thank you both!

PR #1224, opened 2020-10-21
fix FR district scraper

maekke:
- use third row as header
- search for the expected column names instead of expecting the exact name


metaodi:
I really don't know what's going on in FR. They changed that almost everyday since last Friday...

If this continues, I will try to come up with some basic heuristic to find the correct row of the headers.

maekke:
> I really don't know what's going on in FR. They changed that almost everyday since last Friday...
> 
> 
> 
> If this continues, I will try to come up with some basic heuristic to find the correct row of the headers.

Totally agree here...

PR #1226, opened 2020-10-21
use some "safety" whitespace after left axis value detection

maekke:
this fixes this weeks pdf parsing...

### DIFF
--- a/scrapers/scrape_vs_districts.py
+++ b/scrapers/scrape_vs_districts.py
@@ -30,7 +30,7 @@ def strip_left_number(content):
     lines = content.split('\n')
     pos = None
     for line in lines:
-        res = re.search(r'\s+(\d+)', line)
+        res = re.search(r'\s+(\d+)   ', line)
         if res is not None:
             if pos is None:
                 pos = res.end()


scrapers/scrape_vs_districts.py
```diff
@@ -30,7 +30,7 @@ def strip_left_number(content):
     lines = content.split('\n')
     pos = None
     for line in lines:
-        res = re.search(r'\s+(\d+)', line)
+        res = re.search(r'\s+(\d+)   ', line)
```
metaodi, line 33: I'm not sure, but maybe this whitespace should be `\s*`, so that it could be there or not. Or does this break something else?

maekke: I'm not really sure what's best to do here. The "problem" with this weeks data is that content before calling `strip_left_number` looks like that (empty lines removed):
```
                          100         158.1
                                                                                                                                                                                                                 Inciden                                                                                                                                                                                                    200.0
                             50                                                                                                                                                                     100.0
                                          7          33          78          70          27        385          46         293         215         378          70          80        208
                               0                                                                                                                                                                    0.0
```
The 100, 50 and 0 values are not indented identically, which causes the trouble that the scraper fails. Adding the two spaces just works around this whitespace issue - I don't really have a better idea on how to detect and remove the left axis values...
(The complete scraper logic not very nice to be honest, but I couldn't find any better way how to parse the available PDF)

PR #1252, opened 2020-10-26
AG: Remove no longer published data from scraping

Qsi1984:
The information about risk-area-travelers is not published in the excel source anymore. Scraper will fail because the sheet is missing.


metaodi:
I hope the data comes back up (we will contact AG to ask). But in the meantime: thanks for your contribution.

PR #1260, opened 2020-10-27
SZ: fix quarantine_riskareatravel scraped value index

maekke:
fix issue introduced in PR #1259


scrapers/scrape_sz.py
```diff
@@ -28,7 +28,7 @@
     dd.isolated = res[1]
     dd.hospitalized = res[2]
     dd.quarantined = res[3]
-    dd.quarantine_riskareatravel = res[3]
+    dd.quarantine_riskareatravel = res[4]
```
metaodi, line 31: Maybe for those more complex regexes it would make sense to use [named groups](https://docs.python.org/3/howto/regex.html#non-capturing-and-named-groups), e.g.

```python
res = re.search(r'.*\s+(?P<iso>\d+)\s+\d+\s+\d+\s+(?P<hosp>\d+)\s+(?P<quar>\d+)\s+(?P<risk>\d+)\s+', pdf_content)
dd.isolated = res['iso']
dd.hospitalized = res['hosp']
dd.quarantined = res['quar']
dd.quarantine_riskareatravel = res['risk']
```

WDYT?

maekke: Nice, didn't know about that... Should be addressed in 5c24ea6.

Issue #1026, opened 2020-08-31
New file for number of tests per day for Basel-Stadt


jb3-2:
We publish weekly numbers of tests (total, negative, positive) for people living in Basel-Stadt: https://data.bs.ch/explore/dataset/100094 - @metaodi Can you please scrape those and add a new CSV file to the repo? Thanks!

metaodi:
We will define a structure for this data and then start to integrate your data in this repo.

andreasamsler:
@jb3-2 to get it right: BS is publishing once per week the daily numbers of tests (total, negative, positive), correct?

jb3-2:
@andreasamsler Correct! We get the data each Friday and put them online Friday afternoon.

metaodi:
It's finally done: #1342

The file is here: https://github.com/openZH/covid_19/blob/master/fallzahlen_tests/fallzahlen_kanton_BS_tests.csv

PR #1272, opened 2020-10-27
GL: get number of performed PCR tests from weekly PDF

maekke:
Not sure how useful this number is, since almost no cantons provide this value...

### DIFF
--- a/scrapers/scrape_gl.py
+++ b/scrapers/scrape_gl.py
@@ -24,6 +24,16 @@ content = sc.pdftotext(pdf, page=1)
 pdf_date = sc.find(r'Stand: (\d{2}\.\d{2}.\d{4})', content)
 pdf_date = sc.date_from_text(pdf_date)
 
+number_of_tests = sc.find(r'PCR-Tests\sKanton Glarus\s(\d+\'?\d+)\s', content).replace('\'', '')
+is_first = True
+if number_of_tests:
+        dd = sc.DayData(canton='GL', url=pdf_url)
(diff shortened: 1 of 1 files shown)


metaodi:
Do they provide the number of positive and the total number of tests? If so, it would be similar to https://github.com/openZH/covid_19/blob/master/fallzahlen_kanton_zh/COVID19_Anteil_positiver_Test_pro_KW.csv and #1026 

Most cantons get this kind of data from the BAG, some cantons like ZH and BS publish it already. I was thinking of creating a new directory here in the repository (similar to the district data) to collect this data (see #1026)

Not sure if this applies to this PR.

maekke:
> Do they provide the number of positive and the total number of tests? If so, it would be similar to https://github.com/openZH/covid_19/blob/master/fallzahlen_kanton_zh/COVID19_Anteil_positiver_Test_pro_KW.csv and #1026

Unfortunately, they only seem to provide the total amount of PCR tests (total, last 7 and 14 days) and "Positivitätsrate".

> 
> Most cantons get this kind of data from the BAG, some cantons like ZH and BS publish it already. I was thinking of creating a new directory here in the repository (similar to the district data) to collect this data (see #1026)
> 
> Not sure if this applies to this PR.

I guess this wouldn't really work for the available data from GL.

But I like the idea to add another directory with this data, for example TG provides this on a weekly basis as well.

metaodi:
okay then let's merge these two PRs and see if we can find this data for other cantons in a similar format to add to an additional directory.

PR #1280, opened 2020-10-28
add scrapers for performed tests

maekke:
Still a bit work in progress, since it's not really wired up anywhere.

The csv is based on the `COVID19_Anteil_positiver_Test_pro_KW.csv` ZH file, with some additional columns. Is that fine or should it stick to the ZH ones?

Also added an AG common file to retrieve the xlsx url in one place only.
(I guess that might be something for the FR scrapers to do as well)

### DIFF
--- a/scrapers/scrape_common.py
+++ b/scrapers/scrape_common.py
@@ -345,7 +345,7 @@ def download_file(url, path):
         for chunk in r.iter_content(1024):
             f.write(chunk)
 
-def parse_xls(book, header_row=1, sheet_index=0, sheet_name=None, skip_rows=1, columns_to_parse=None):
+def parse_xls(book, header_row=1, sheet_index=0, sheet_name=None, skip_rows=1, columns_to_parse=None, enable_float=False):
     rows = []
     if sheet_name:
         sheet = book.sheet_by_name(sheet_name)
@@ -367,9 +367,9 @@ def parse_xls(book, header_row=1, sheet_index=0, sheet_name=None, skip_rows=1, c
                 entry[h] = xlrd.xldate.xldate_as_datetime(value, book.datemode)
             elif cell_type == xlrd.XL_CELL_EMPTY:
                 entry[h] = None
-            elif represents_int(value) and int(value) == value:
+            elif represents_int(value):
                 entry[h] = int(value)
-            elif represents_float(value):
+            elif enable_float and represents_float(value):
                 entry[h] = float(value)
             else:
                 entry[h] = value
--- a/scrapers/scrape_common.py
+++ b/scrapers/scrape_common.py
@@ -190,6 +190,76 @@ class DistrictData:
         return 'DistrictId,District,Canton,Date,Week,Year,Population,TotalConfCases,NewConfCases,TotalDeaths,NewDeaths,SourceUrl'
 
 
+class TestData:
+    __initialized = False
+    SEPARATOR = ','
+
+    def __init__(self, canton=None, url=None):
+        self.start_date = None
+        self.end_date = None
+        self.week = None
+        self.year = None
+        self.canton = canton
+        self.positive_tests = None
+        self.negative_tests = None
+        self.positivity_rate = None
+        self.url = url
+        self.__initialized = True
+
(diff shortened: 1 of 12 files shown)


metaodi:
@andreasamsler for #1026 we talked about creating a structure for those data about tests. Should we stick to the ZH file or is there already a structure (e.g. name or variables) defined somewhere?

maekke:
while at it, added scrapers for SG and FR that provide complete data.

scrapers/scrape_common.py
```diff
@@ -190,6 +190,76 @@ def header():
...
+        res.append(self.canton)
+        res.append('' if self.start_date is None else str(self.start_date))
+        res.append('' if self.end_date is None else str(self.end_date))
+        res.append('' if self.week is None else str(self.week))
+        res.append('' if self.year is None else str(self.year))
+        res.append('' if self.positive_tests is None else str(self.positive_tests))
+        res.append('' if self.negative_tests is None else str(self.negative_tests))
+        res.append('' if self.positivity_rate is None else str(self.positivity_rate))
+        res.append(self.url)
+        return DistrictData.SEPARATOR.join(res)
```
metaodi, line 225: This should be `TestData.SEPARATOR`

scrapers/scrape_common.py
```diff
@@ -190,6 +190,76 @@ def header():
...
+    @staticmethod
+    def __get_int_item(item):
+        return int(item) if represents_int(item) else None
+
+    @staticmethod
+    def __get_float_item(item):
+        return float(item) if represents_float(item) else None
+
+    def parse(self, data):
+        items = data.split(DistrictData.SEPARATOR)
```
metaodi, line 244: Dito

scrapers/scrape_common.py
```diff
@@ -297,6 +367,8 @@ def parse_xls(book, header_row=1, sheet_index=0, sheet_name=None, skip_rows=1, c
                 entry[h] = xlrd.xldate.xldate_as_datetime(value, book.datemode)
             elif cell_type == xlrd.XL_CELL_EMPTY:
                 entry[h] = None
+            elif support_float and represents_float(value):
```
metaodi, line 370: why do we need this flag? Is there situation, where float should nlt be supported?

metaodi (changes requested):

scrapers/scrape_common.py
maekke: good point, should be addressed in 73ce2c6

scrapers/scrape_common.py
maekke: good point, should be addressed in 73ce2c6

scrapers/scrape_common.py
maekke: Each `int` value is also a `float` value, that's why I added the flag and `represents_float` function.
I hope to have improved this in ffb3ca7: the `represents_int` is extended to verify that the value is actually an integer and if that's not the case, it's being checked for `float`.
Do you think this makes more sense?

maekke:
Also added BS #1026

maekke:
The format sticks to the ZH format right now. I think adding a _Total Tests_ column would be very reasonable, since some cantons provide this number and the positivity rate. So we could just use these values and not do any assumptions when calculating number of positive and negative tests.
For those cantons providing it weekly, maybe it would also make sense to add the week/year columns (but that's strictly not needed).

@metaodi are you fine with the change about splitting out some common cantonal scraper stuff into for example scraper_ag_common.py?
I could cherry-pick them into separate PRs, so this focuses on the performed test scraping.

And I guess the naming is also open, I personally prefer the things in English...

And maybe something completely different: what about the new "Antigen Tests"? Do you know if this will be provided as separate numbers or merged together?

metaodi:
Yes I like the cantonal "common" modules, let's keep them.

I guess we will switch to English names, since the [latest new files](https://github.com/openZH/covid_19/blob/master/fallzahlen_kanton_zh/COVID19_Belegung_Intensivpflege.csv) follow this pattern as well.

I reminded @andreasamsler today, that this is still a pending issue.

PR #1281, opened 2020-10-29
Handle typo in case number for JU on 2020-05-28

metaodi:


metaodi:
JU fixed their source, so this PR is no longer needed.

PR #1304, opened 2020-11-01
VS: scrape daily PDF

maekke:
the [latest PDF](https://www.vs.ch/documents/6756452/7008787/2020%2010%2030%20-%20Sit%20Epid%20-%20%C3%89tat%20Stand) contains more recent data than the excel file, might be worth scraping this file...

### DIFF
--- a/.github/workflows/run_scrapers.yml
+++ b/.github/workflows/run_scrapers.yml
@@ -61,7 +61,7 @@ jobs:
         pip install -r requirements.txt
         sudo apt update || true # do not fail if update does not work
         sudo apt-get install sqlite3
-        if [ "$SCRAPER_KEY" = "FL" ] || [ "$SCRAPER_KEY" = "GE" ] || [ "$SCRAPER_KEY" = "VD" ] || [ "$SCRAPER_KEY" = "GL" ] || [ "$SCRAPER_KEY" = "SZ" ] || [ "$SCRAPER_KEY" = "TI" ] || [ "$SCRAPER_KEY" = "SO" ] ; then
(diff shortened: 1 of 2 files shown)


.github/workflows/run_scrapers.yml
```diff
@@ -61,7 +61,7 @@ jobs:
         pip install -r requirements.txt
         sudo apt update || true # do not fail if update does not work
         sudo apt-get install sqlite3
-        if [ "$SCRAPER_KEY" = "FL" ] || [ "$SCRAPER_KEY" = "GE" ] || [ "$SCRAPER_KEY" = "VD" ] || [ "$SCRAPER_KEY" = "GL" ] || [ "$SCRAPER_KEY" = "SZ" ] || [ "$SCRAPER_KEY" = "TI" ] || [ "$SCRAPER_KEY" = "SO" ] ; then
+        if [ "$SCRAPER_KEY" = "FL" ] || [ "$SCRAPER_KEY" = "GE" ] || [ "$SCRAPER_KEY" = "VD" ] || [ "$SCRAPER_KEY" = "GL" ] || [ "$SCRAPER_KEY" = "SZ" ] || [ "$SCRAPER_KEY" = "TI" ] || [ "$SCRAPER_KEY" = "SO" ] || [ "$SCRAPER_KEY" = "VS" ] ; then
```
metaodi, line 64: It's probay time to get rid of this condition andmsimply install it for all cantons.

It made sense when only a couple of cantons needed PDF scraping capabilities.

scrapers/scrape_vs.py
```diff
@@ -37,5 +66,4 @@
         dd.recovered = sum(r['Nb de nouvelles sorties'] for r in rows[:i+1])
     if not is_first:
         print('-' * 10)
-    is_first = False
```
metaodi, line 40: I think this is still needed, should the condition above (line 35) be false, then `is_first` must be set here.

maekke: Whoops, you're right! Reverted this change in d9dcf5c

.github/workflows/run_scrapers.yml
maekke: I thought about that as well, should be addressed in 77f43e3.
Or should we just have one `sudo apt-get install xxx` call with sqlite3 and poppler-utils in one shot?

PR #1305, opened 2020-11-01
GE: scrape current hospitalizations

maekke:
Could be reasonable to get the hospitalized values also on weekends...
While at it, update the number of performed PCR tests regex.

### DIFF
--- a/scrapers/scrape_ge.py
+++ b/scrapers/scrape_ge.py
@@ -20,7 +20,7 @@ pdf = sc.pdfdownload(pdf_url, silent=True)
 
 week_number = sc.find(r'Situation semaine (\d+)', pdf)
 week_end_date = datetime.datetime.strptime('2020-W' + week_number + '-7', '%G-W%V-%u').date()
-number_of_tests = sc.find(r'N total tests : (\d+\'\d+)', pdf)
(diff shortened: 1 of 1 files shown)


scrapers/scrape_ge.py
```diff
@@ -31,6 +31,25 @@
     print(dd_test)
     is_first = False
 
+
+# get hospitalized number
+hosp_url = 'https://www.hug.ch/coronavirus-maladie-covid-19/situation-aux-hug'
```
metaodi, line 36: But this is only one hospital, right? I'm not sure, but I guess there must be others as well.

maekke: I wasn't really sure in the first place. But it seems that it's not a single hospital from the plural name :)
So it's also linked from the [GE covid webpage](https://www.ge.ch/covid-19-situation-epidemiologique-geneve-informations-du-canton/situation-epidemiologique-geneve).
And the values in the [graph](https://www.hug.ch/coronavirus-maladie-covid-19/situation-aux-hug) from 30th of October seem to match with what is [in the repository](https://github.com/openZH/covid_19/blob/master/fallzahlen_kanton_total_csv_v2/COVID19_Fallzahlen_Kanton_GE_total.csv#L282) (458 hospitalizations)

PR #1317, opened 2020-11-02
Separate python requirements for OCR

metaodi:


maekke:
Or should we pin the opencv-python version to 4.4.0.44? [that is the run with 4.4.044](https://github.com/openZH/covid_19/runs/1342913201?check_suite_focus=true#step:9:25) and then the [failing run with 4.4.0.46](https://github.com/openZH/covid_19/runs/1343218654?check_suite_focus=true#step:9:25) ?

maekke:
https://pypi.org/project/opencv-python/#history

metaodi:
Maybe you're right. I'm still not sure, if we should kick out opencv altogether...

PR #1321, opened 2020-11-03
SO: switch to main page data scraping

maekke:
since the table page wasn't updated since 2020-10-30

### DIFF
--- a/scrapers/scrape_so.py
+++ b/scrapers/scrape_so.py
@@ -100,4 +100,3 @@ for row in rows:
         print('-' * 10)
     is_first = False
     print(row)
-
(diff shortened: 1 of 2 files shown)


scrapers/scrape_so.py
```diff
@@ -40,7 +40,8 @@
 
 soup = BeautifulSoup(d, 'html.parser')
 data_table = soup.find('h2', text=re.compile("Situation Kanton Solothurn")).find_next("table")
-if data_table:
+# not updated anymore after 2020-10-30
+if data_table and False:
```
metaodi, line 44: I would remove the `and False` part, as it's not really a problem if this is scraped as well. Maybe at some point the table is updated. If we later think the table is deprecated, I'd rather remove the whole code block.

maekke: That's a good point, should be addressed with adfd0b9 to scrape both the table and main page.

PR #1342, opened 2020-11-08
add performed tests runner

maekke:
add the missing pieces to get performed tests running

- changed the csv file names to use English names
- moved a few pieces to a new db_common.py file
- added a 'total tests' column for the cantons providing only this number and the positivity rate

### DIFF
--- a/.github/workflows/run_tests_scraper.yml
+++ b/.github/workflows/run_tests_scraper.yml
@@ -1,4 +1,4 @@
-name: Run district scrapers
+name: Run tests scrapers
 
 on:
   schedule:
--- a/.github/workflows/run_tests_scraper.yml
+++ b/.github/workflows/run_tests_scraper.yml
@@ -1,0 +1,81 @@
+name: Run district scrapers
+
+on:
+  schedule:
+  - cron:  '20 * * * *' # run every hour at xx:20
+  workflow_dispatch: ~
+jobs:
+  run_scraper:
+    runs-on: ubuntu-16.04
+    continue-on-error: false
+    timeout-minutes: 10
+    strategy:
+      fail-fast: false
+      matrix:
+        canton:
+          - AG
+          - BE
+          - BS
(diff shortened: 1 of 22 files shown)


.github/workflows/run_tests_scraper.yml
```diff
@@ -0,0 +1,81 @@
+name: Run district scrapers
```
metaodi, line 1: This should be something like "Run tests scraper"

maekke: You're right! Should be fixed in https://github.com/openZH/covid_19/pull/1342/commits/f7b01d50ef3f5580f657dc40efc69ec2b5d17c65

metaodi:
I just had a phone call with @andreasamsler to discuss this PR

Notes:

- positivity_rate is missing from the CSV
- if possible we should not to calculations on our own, but rather provide the data by the cantons (so either have `positive_tests` and `negative_tests` OR have `total_tests` and `positivity_rate`).  Depending on what numbers the cantons provide.
- The field `url` should be renamed to `source` to be consistent with the other CSVs

Other than that, this is good to go!

maekke:
> I just had a phone call with @andreasamsler to discuss this PR
> 
> Notes:
> 
>     * positivity_rate is missing from the CSV
> 
>     * if possible we should not to calculations on our own, but rather provide the data by the cantons (so either have `positive_tests` and `negative_tests` OR have `total_tests` and `positivity_rate`).  Depending on what numbers the cantons provide.
> 
>     * The field `url` should be renamed to `source` to be consistent with the other CSVs
> 
> 
> Other than that, this is good to go!

Alright, we're getting closer then!

Thanks for spotting the missing positivity_rate, should be added. Also renamed the url to source.

I've updated all scrapers to avoid any calculations, but add what is available from the source.

metaodi:
Thanks a lot @maekke !

PR #1352, opened 2020-11-09
adjust AG scraper to work with latest contact tracing data

maekke:
not sure if there is a better way...


metaodi:
I just changed that today, so I will just switch to what it was before: #1354

For AG I added a clause to the `parse_xls` method to add the column number to the header name if there are duplicate column names.

PR #1359, opened 2020-11-11
update LU scraper to use different csv files as historized sources

### LINKED ISSUE #1358: LU scraping URL changed
currently: [https://www.lustat.ch/analysen/gesundheit/corona-reporting](https://www.lustat.ch/analysen/gesundheit/corona-reporting)
with different page layout

maekke:
could fix #1358 if it actually works...

### DIFF
--- a/scrapers/scrape_lu.py
+++ b/scrapers/scrape_lu.py
@@ -1,9 +1,71 @@
 #!/usr/bin/env python3
 
+import csv
+from io import StringIO
 import re
 import datetime
 from bs4 import BeautifulSoup
 import scrape_common as sc
+
+
+def fix_lu_date(date):
+    res = re.match(r'(20\d{2})/(\d+)/(\d+)', date)
+    assert res, 'date could not be matched!'
+    date = datetime.date(int(res[1]), int(res[2]) + 1, int(res[3]))
+    return date.isoformat()
+
+
+hosp_url = 'https://www.lustat.ch/analysen/gesundheit/corona-reporting/hospitalisationen'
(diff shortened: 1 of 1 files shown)


metaodi:
I'm using this PR as a starting point, I don't think we get better data any time soon.

maekke:
@metaodi I added scraping of isolated, quarantined and travel quarantined in f2b0e63 and removed the old logic. I guess that should be all, what's needed?

metaodi:
@maekke you mentioned something about the dates not matching or being off by one (can't remember exactly). That's why I wanted to check if the out put somewhat matches with what we have in the repo.

But yes I agree, that's all!

maekke:
> @maekke you mentioned something about the dates not matching or being off by one (can't remember exactly). That's why I wanted to check if the out put somewhat matches with what we have in the repo.
> 
> But yes I agree, that's all!

Ah I see, the month "off by one" is taken care of in `fix_lu_date`. So from my side it should be fine as well!

metaodi:
> > @maekke you mentioned something about the dates not matching or being off by one (can't remember exactly). That's why I wanted to check if the out put somewhat matches with what we have in the repo.
> > But yes I agree, that's all!
> 
> Ah I see, the month "off by one" is taken care of in `fix_lu_date`. So from my side it should be fine as well!

btw: in the meantime I figured out why the month is off by one: these CSVs are used to display tHe Highcharts charts using JavaScript. And in JavaScript the months actually start at 0, i.e. in JS `Date.UTC(2020, 10, 12)` is equal to 2020-11-12.

PR #1372, opened 2020-11-13
fix parent element for NW scraper

maekke:


metaodi:
I already changed it today from `h3` to `h4`. Now they changed it again to a `p`. If this keeps happening we need to make this independent of the tag they use...

PR #1396, opened 2020-11-18
FR: use German website to get the xls url from

maekke:


maekke:
FR page was fixed as is seems. Can also be closed, I don't mind...

metaodi:
Thanks anyway!

PR #1399, opened 2020-11-18
Add FL tests scraper

maekke:


scrapers/scrape_fl_tests.py
```diff
@@ -0,0 +1,17 @@
...
+url = 'https://www.llv.li/files/as/grafik_covid19_tests_pro_kw.xlsx'
+xls = sc.xlsdownload(url, silent=True)
+rows = sc.parse_xls(xls, header_row=52, sheet_name='gTests_AG')
+for row in rows:
+    if row['B'] is None:
+        # skip the footer line
+        continue
+    td = sc.TestData(canton='FL', url=url)
+    td.week = sc.find(r'KW (\d+)', row['B'])
+    td.year = '2020'
```
metaodi, line 14: I already fear for next year, when we have to check the codebase for things like that. Maybe they will provide a year column until then...

maekke: I agree, that this is far from optimal. The excel file contains some dates, but I think it's not straight forward to guess it from it. I would also hope, that a year will be added next month or so...

PR #1423, opened 2020-11-26
AR: update death regex

maekke:


metaodi:
I had to do a follow-up PR (#1427) of this, since we suddenly had 1442 deaths for AR... 😲

maekke:
> I had to do a follow-up PR (#1427) of this, since we suddenly had 1442 deaths for AR... 😲

Whoops, Merci!

PR #1433, opened 2020-11-27
update FL scraper death regex once again

maekke:
hope that we don't need to update this regex every day...


metaodi:
> hope that we don't need to update this regex every day...

The joy of parsing natural language with regex 😅

PR #1442, opened 2020-11-30
Add VD tests scraper

maekke:
the provided value of average number of tests per day doesn't really fit I guess, so there is a very basic multiplication - I hope this is fine?
Otherwise this could be done on a per day basis with the provided value and positivity rate for the week dates (which is also not very nice I guess).

### DIFF
--- a/scrapers/scrape_vd_tests.py
+++ b/scrapers/scrape_vd_tests.py
@@ -17,19 +17,21 @@ for pdf_url in pdf_urls:
     year = sc.find(r'Situation au \d+.*(20\d{2})', pdf)
     res = re.search(r'Entre\s+le\s+(\d+\s+\w+)\s+et\s+le\s+(\d+\s+\w+),', pdf)
     if res:
-        td.start_date = sc.date_from_text(f'{res[1]} {year}').isoformat()
-        td.end_date = sc.date_from_text(f'{res[2]} {year}').isoformat()
+        start_date = sc.date_from_text(f'{res[1]} {year}')
+        end_date = sc.date_from_text(f'{res[2]} {year}')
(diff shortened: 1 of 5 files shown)


scrapers/scrape_vd_tests.py
```diff
@@ -0,0 +1,38 @@
...
+        res = re.search(r'Entre\s+le\s+(\d+)\s+et\s+le\s+(\d+\s+\w+),', pdf)
+        if res:
+            end_date = sc.date_from_text(f'{res[2]} {year}')
+            td.end_date = end_date.isoformat()
+            td.start_date = sc.date_from_text(f'{res[1]}.{end_date.month}.{year}').isoformat()
+    assert td.start_date and td.end_date, f'failed to extract start and end dates from {pdf_url}'
+
+    res = re.search(r'une\s+moyenne\s+de\s+(\d+)\s+frottis\s+SARS-CoV(-)?2', pdf)
+    assert res, f'failed to extract total number of tests from {pdf_url}'
+    td.total_tests = 7 * int(res[1])
```
metaodi, line 32: As you mentioned, we only have a weekly average, so I think this is okay. But maybe it's better to calculate the difference between the `start_date` and `end_date` to be safe (instead of hard-coding `7`). WDYT?

maekke: Sounds very reasonable! I've updated the PR (and also fixed the order for getting the PDFs, it should pick the latest ones available).

metaodi:
@maekke currently GitHub Actions has some issues (see https://www.githubstatus.com/). So it may take some time to run all the checks and the scrapers.

PR #1452, opened 2020-12-03
Restore SO main page scraping

maekke:
@metaodi was there a good reason to remove main page scraping? It seems to contain more up-to-date data than the other ones...


metaodi:
@maekke the scraper kept failing on the main page and the table seemed to have the same data. But looking at it now it seems justified to scrape both. And thanks to the updated header texts, all values should be there again. So thanks for fixing this!

PR #1458, opened 2020-12-05
JU: sanitize table to skip unavailable values

maekke:
@metaodi not sure if there is a better way?

### DIFF
--- a/scrapers/scrape_ju.py
+++ b/scrapers/scrape_ju.py
@@ -16,6 +16,7 @@ def sanitize_row(row):
         row['Nombre de cas actuellement en soins intensifs'] = ''
     if not sc.represents_int(row.get('Nombre de nouveaux décès')):
         row['Nombre de nouveaux décès'] = ''
+    return row
 
 
 url = 'https://www.jura.ch/fr/Autorites/Coronavirus/Chiffres-H-JU/Evolution-des-cas-COVID-19-dans-le-Jura.html'
@@ -46,7 +47,7 @@ if data_table:
                 print('-' * 10)
             is_first = False
 
-            sanitize_row(row)
+            row = sanitize_row(row)
 
             dd = sc.DayData(canton='JU', url=url)
             current_year = datetime.datetime.now().year
--- a/scrapers/scrape_ju.py
+++ b/scrapers/scrape_ju.py
@@ -26,6 +26,18 @@ if data_table:
         rows.append(data)
 
     if rows:
+        # sanitize data:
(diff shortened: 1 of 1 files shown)


scrapers/scrape_ju.py
```diff
@@ -26,6 +26,18 @@
...
+        for row in rows:
+            if not row.get('Date') or row.get('Date') == 'Date':
+                continue
+            if not sc.int_or_word(row.get('Nombre de cas actuellement hospitalisés')):
+                row['Nombre de cas actuellement hospitalisés'] = ''
+            if not sc.int_or_word(row.get('Nombre de cas actuellement en soins intensifs')):
+                row['Nombre de cas actuellement en soins intensifs'] = ''
+            if not sc.int_or_word(row.get('Nombre de nouveaux décès')):
+                row['Nombre de nouveaux décès'] = ''
+
```
metaodi, line 40: Instead of creating an extra for loop, I would add a method called something like `sanitize_row` where you could make those checks.
And instead of using `sc.int_or_word` I would use `sc.represents_int`, which should cover these cases.

metaodi (changes requested):

scrapers/scrape_ju.py
maekke: Good point, should be addressed in c0597ee.

scrapers/scrape_ju.py
```diff
@@ -6,6 +6,18 @@
...
+def sanitize_row(row):
+    # sanitize data:
+    # 2020-12-04 contains 'Non communiqué' entries, skip them for now
+    if not sc.represents_int(row.get('Nombre de cas actuellement hospitalisés')):
+        row['Nombre de cas actuellement hospitalisés'] = ''
+    if not sc.represents_int(row.get('Nombre de cas actuellement en soins intensifs')):
+        row['Nombre de cas actuellement en soins intensifs'] = ''
+    if not sc.represents_int(row.get('Nombre de nouveaux décès')):
+        row['Nombre de nouveaux décès'] = ''
+
```
metaodi, line 19: You should return the row here

scrapers/scrape_ju.py
```diff
@@ -34,6 +46,8 @@
                 print('-' * 10)
             is_first = False
 
+            sanitize_row(row)
```
metaodi, line 49: ...and assign it here:

```python
row = sanitize_row(row)
```

maekke: Should both be addressed in 18f6a3e986a89fb4135b34711fe03b1e6ff0f9fe

metaodi (approved):

PR #1459, opened 2020-12-05
Fix FL table PDF scraping

maekke:
Values newer than 2020-11-16 are not considered, due to `1’002` values not being matched by the regex.


metaodi (approved):
LGTM

PR #1460, opened 2020-12-05
Update GE scrapers

maekke:
There is a new page available: https://infocovid.smc.unige.ch/

Get isolated and quarantined data and performed tests from there.


maekke:
this doesn't work yet, I guess something is wrong with the session or so.

PR #1461, opened 2020-12-06
JU: fix all table entries before extracting the data

maekke:
@metaodi I guess it's better to fix all the entries in the table first... :shrug:

### DIFF
--- a/scrapers/scrape_ju.py
+++ b/scrapers/scrape_ju.py
@@ -39,6 +39,9 @@ if data_table:
         rows.append(data)
 
     if rows:
+        for row in rows[:-1]:
+            row = sanitize_row(row)
+
         for i, row in enumerate(rows[:-1]):
             if not row.get('Date') or row.get('Date') == 'Date':
                 continue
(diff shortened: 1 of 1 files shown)


scrapers/scrape_ju.py
```diff
@@ -39,6 +39,9 @@ def sanitize_row(row):
         rows.append(data)
 
     if rows:
+        for row in rows[:-1]:
+            row = sanitize_row(row)
+
```
metaodi, line 44: I'm no expert in this, but I think if you modify the rows they are not changed in-place in `rows`, so iterating again over `rows` does not result in having the sanitized values. So I guess you have to create a new list `sanitized_rows` and append the sanitized rows to it. Or am I missing something?

maekke: The `row` object is a dictionary and thus it will be modified by the `sanitize_row` function in the original list already.
There is a chapter in the [Python FAQs](https://docs.python.org/3/faq/programming.html#how-do-i-write-a-function-with-output-parameters-call-by-reference) that explains the behaviour. Basically the primitive types like `int` and `str`ings get copied as value, but almost everything else is mutable and thus is passed as a "reference" and modifies the source item. ([The first paragraph(s) explain it also not too bad](https://www.oreilly.com/library/view/learning-python/1565924649/ch04s04.html))

metaodi: Alright, thanks for the pointers!

PR #1465, opened 2020-12-07
update GE tests scraper to use page 4 and 6 for PCR test data

maekke:


maekke:
@metaodi do you think this is fine?
a more reasonable approach could be to go through all the pages and check if the expected title is found and parse the page then...

metaodi:
> @metaodi do you think this is fine?
> a more reasonable approach could be to go through all the pages and check if the expected title is found and parse the page then...

It's just very likely that this code must be adapted later. Especially for GE, which changes it's documents quite regularly. I'm fine with merging it like that, because GE will always require special care 😉

metaodi (approved):

maekke:
> > @metaodi do you think this is fine?
> > a more reasonable approach could be to go through all the pages and check if the expected title is found and parse the page then...
> 
> It's just very likely that this code must be adapted later. Especially for GE, which changes it's documents quite regularly. I'm fine with merging it like that, because GE will always require special care wink

Sure, I've updated the PR. I hope that this robust enough for now...

PR #1471, opened 2020-12-11
pin xlrd version to 1.2.0

maekke:
this should fix the most recent failures... it looks like something behaves different with version 2.0

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 beautifulsoup4
 requests
 dateparser
-xlrd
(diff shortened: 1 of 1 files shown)


fabian:
Version 2.0.0 removed support for anything else than `.xls`, so `.xlsx` is no longer supported: https://github.com/python-excel/xlrd/commit/c6d642b4a2688df5855b06ebb0403f1806805fbb

metaodi:
Thank you @maekke

PR #1495, opened 2020-12-17
AR: use the same date/time regex for contact tracing as for the cases

maekke:


maekke:
whoops, sorry I accidentally deleted the wrong branch...

PR #1496, opened 2020-12-19
Fix BL bulletin parsing

maekke:
it looks like the quarantined numbers from the last few weeks for BL are wrong. I'll try to manually fix this in the data after this is merged.

The changes should fix this for both the normal scraper and tests scraper.


maekke:
Previous values should be fixed in the csv in https://github.com/openZH/covid_19/pull/1496/commits/cb6fb1fafd133609105f21742efc99d6d02b43b4

PR #1498, opened 2020-12-19
"fix" JU scraper to avoid accessing inexistant table

maekke:
not really sure if this should be merged or not... maybe it's better to have a failing scraper, when the data is not there :man_shrugging:


metaodi:
I think it should fail. While we wait for JU to answer, I temporarly disabled their scraper.

PR #1504, opened 2020-12-22
Update BS scraper

metaodi:


maekke:
@metaodi LGTM, I've added one question about the source.

maekke (approved):

PR #1530, opened 2020-12-30
Skip ZG deaths on 2020-12-29

maekke:


maekke:
ZG fixed the source, closing.

PR #1538, opened 2021-01-01
remove unneeded year logic from JU scraper

maekke:
that causes problems to output the date with 2020 and 2021 afterwards


metaodi:
Wow, Jura added the year to the date column! 🎉🎉🎉

PR #1539, opened 2021-01-01
remove confirmed cases from AG for now

maekke:
no data available on 2020-12-31


maekke:
No longer required...

metaodi:
I guess this is no longer needed. I'd rather keep the check unless we're sure it's not always delivered.

maekke:
agreed, let's close this one.

PR #1554, opened 2021-01-05
update PDF matching regex for FL

maekke:
it looks like the last three PDFs cannot really be parsed (the first 3 look fine), the produced text is super weird... Not sure if it's worth trying update the regex for it (with `layout=True` things are still a bit broken).

### DIFF
--- a/scrapers/scrape_fl.py
+++ b/scrapers/scrape_fl.py
@@ -9,7 +9,7 @@ import re
 d = sc.download('https://www.regierung.li/coronavirus', silent=True)
 
 is_first = True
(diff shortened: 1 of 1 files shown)


metaodi:
We can also ditch this PDF and only use the second one with the list. WDYT @maekke ?
By using the "Situationsbericht" we are a couple hours or maybe a day ahead with the numbers, but this is not really worth it IMHO.

maekke:
@metaodi I've changed the scraper to use the daily HTML items instead of the PDF files, see d5b9483.

metaodi (approved):
LGTM

PR #1556, opened 2021-01-06
Update date regex for VD tests

metaodi:
In the latest PDF the date is written like this:

`Entre le 28 décembre 2020 et le 4 janvier 2021,` so a new regex is needed to parse the year.

And the UNICODE flag is needed, so that python recognizes the `é` in décembre as being a word-character (matched with `\w`)


maekke:
LGTM, thanks! :+1:

PR #1573, opened 2021-01-10
use numeric ordering for weeks for tests/district csv output

maekke:
https://github.com/openZH/covid_19/commit/411e64b1705edb282f0f5f8874e28263eac16993 shows that the weeks are ordered as string instead of a numerical ordering. This should be fixed with the `+0` (to cast it to a number), when creating the csv from the sqlite db.

### DIFF
--- a/scrapers/run_district_scraper.sh
+++ b/scrapers/run_district_scraper.sh
@@ -30,5 +30,5 @@ $scrape_script | $DIR/add_district_db_entry.py
 
 # 3. Export the database as csv
 echo "Export database to CSV..."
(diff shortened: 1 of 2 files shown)


metaodi:
Should we define week as a number in the sqlite db?

maekke:
> Should we define week as a number in the sqlite db?

That could work as well, but I thought that this would be problematic, because the field is `NOT NULL`, but kind of "optional" (in case the start/end date is used). But it seems to work, when changing it to `integer NOT NULL`.
The same fix should be applied to the districts scraper too.
I'll have another look at this...

maekke:
@metaodi I have the feeling that this fix works better than using a `week integer NULL` in the DB. At least for the district, this created duplicate entries... If you have a better idea, feel free :sweat_smile: 
I've adjusted the district scraper script the same way: https://github.com/openZH/covid_19/pull/1573/commits/bf36ca28c63915b24353d0ed3e43688fc3d9db16

Issue #1587, opened 2021-01-13
New page Bern (districts)


balzamas:
Bern revamped their Corona page, new URL for district data is:
https://www.besondere-lage.sites.be.ch/de/start/news/fallzahlen.html

maekke:
Thanks for reporting David!
PR for the scraper and tests scraper is #1588
PR for the district scraper update is at #1591

maekke:
both merged.

PR #1591, opened 2021-01-13
use new website for BE district scraper

maekke:
- make sure to pick the correct year
- update to new structure
- skip 'Unbekannt' items


maekke:
Resolves #1587

PR #1596, opened 2021-01-16
use main page as source

maekke:
for the scrapers / cantons that update the excel URL for new data.
I'd prefer changing this, since this would cause less changes in the csv files and thus would be easier to track, what actually changed. Of course this is a bit a change of the behaviour...


metaodi:
The frequent changes do not really bother me, but I'm fine with changing to a more stable URL.

PR #1615, opened 2021-01-26
GL: Make sure dates in december are in 2020

metaodi:


metaodi:
@maekke it's ugly, but it only has to work another year...

maekke:
@metaodi I guess there is no other way... thanks for fixing it!

PR #1617, opened 2021-01-26
SH: add one more regex for positivity rate to match this weeks PDF

maekke:


metaodi:
Oh I didn't check the PR list and created #1618 , but your solution is more elegant.

PR #1618, opened 2021-01-27
Add another regex to match the positivity rate in SH

metaodi:


metaodi:
Closing in favour of #1617

PR #1631, opened 2021-02-02
UR: fix date regex for todays typo

maekke:


metaodi:
That was OW not UR ;)

PR #1644, opened 2021-02-03
VD: add optional year to main regex and remove unicode

maekke:
not really sure why `UNICODE` is an issue here (due to the février maybe?). But it seems to work...


metaodi:
@maekke the UNICODE flag ensures that non-ascii characters are recognized as `\w` (regex word character).

metaodi:
Is there any downside to use `re.I|re.UNICODE` ?

maekke:
> Is there any downside to use `re.I|re.UNICODE` ?

@metaodi I've updated the PR and restored `re.UNICODE`. I guess there was another issue with the regex itself...

PR #1657, opened 2021-02-10
Use wait method of selenium

metaodi:


maekke:
@metaodi cool, thanks for improving this! :+1:

PR #1661, opened 2021-02-12
GE: wait for link to be present

maekke:
@metaodi I think the last change was not really working well. This one seems to work at least locally.


metaodi:
@maekke thank you! I had troubles to setup selenium locally so I was never able to test this properly.

PR #1680, opened 2021-02-21
add a timeout of 5s to download function

maekke:
Kind of a follow up of #1676 to add an explicit timeout. I'm not fully sure what the default is, but maybe this helps to avoid the few scraper errors.


metaodi:
I will additionally add `backoff_factor=2` since I realized this is set to 0 by default (i.e. all retries are executed without waiting in between)

PR #1685, opened 2021-02-21
fix(tg-districts): remove wrong data from TG districts

dr-itz:


metaodi:
@dr-itz can you elaborate on this PR? How is the data wrong?

metaodi:
Ah nevermind, the week numbers are wrong, got it.

PR #1817, opened 2021-06-07
Restore BE district scraper

### LINKED ISSUE #1816: New format Bern District numbers
Unfortunately only Incidence numbers and only from the last week, older ones are not in the csv:
https://raw.githubusercontent.com/openDataBE/covid19Data/develop/7_d_inzidenz_verwaltungskreis.csv

Any way to get this in a way so it is consistent with the existing data?

maekke:
@metaodi the computation from the "7 Tage Inzidenz" back to the absolute number of cases seem to produce pretty reasonable results (for example Bern-Mittelland ` 23.43 / 100000 * = 96.9809874`)

Resolves #1816

### DIFF
--- a/scrapers/scrape_be_districts.py
+++ b/scrapers/scrape_be_districts.py
@@ -35,5 +35,5 @@ for row in reader:
     week = date.isocalendar()[1]
     dd.week = week
     dd.year = date.year
-    dd.new_cases = round(float(row['7_d_inzidenz']) / 100e3 * float(row['einwohnerzahl']))
+    dd.new_cases = round(float(row['7_d_inzidenz']) / 100e3 * int(row['einwohnerzahl']))
     print(dd)
--- a/scrapers/scrape_be_districts.py
+++ b/scrapers/scrape_be_districts.py
@@ -1,79 +1,39 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 
-import re
-from bs4 import BeautifulSoup
+import csv
+from io import StringIO
 import scrape_common as sc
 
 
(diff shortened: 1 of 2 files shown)


scrapers/scrape_be_districts.py
```diff
@@ -1,79 +1,39 @@
...
+    district_id = int(row['bfs_nummer'])
+    dd = sc.DistrictData(district=district_ids[district_id], canton='BE')
+    dd.url = url
+    dd.district_id = district_id
+    dd.population = row['einwohnerzahl']
+    date = sc.date_from_text(row['datum'])
+    week = date.isocalendar()[1]
+    dd.week = week
+    dd.year = date.year
+    dd.new_cases = round(float(row['7_d_inzidenz']) / 100e3 * float(row['einwohnerzahl']))
```
metaodi, line 38:
```suggestion
    dd.new_cases = round(float(row['7_d_inzidenz']) / 100e3 * int(row['einwohnerzahl']))
```

The result of this should be an integer, right? Or is there something I'm missing?

maekke: Yep, it's an integer - you are of course right! I've applied your suggestion, thank you.

balzamas:
It seems that the data is updated daily now, not weekly. 
https://github.com/openZH/covid_19/blob/master/fallzahlen_bezirke/fallzahlen_kanton_BE_bezirk.csv has now week 23 even though the week just begun. I checked the history of the source csv from  @openDataBE and it indeed is a daily update.

metaodi:
@balzamas I created a new PR #1819 to address this.

PR #1819, opened 2021-06-09
BE: save daily values

metaodi:
Instead of weekly values, switch back to daily values.

Based on this feedback: https://github.com/openZH/covid_19/pull/1817#issuecomment-857398866

### DIFF
--- a/scrapers/scrape_be_districts.py
+++ b/scrapers/scrape_be_districts.py
@@ -31,6 +31,6 @@ for row in reader:
     dd.url = url
     dd.district_id = district_id
     dd.population = row['einwohnerzahl']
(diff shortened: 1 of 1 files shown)


balzamas:
It seems pretty tricky, as there is only the incidence rate, quite difficult to calculate the daily new cases from this. Looks impossible to me.

maekke:
@metaodi I'm a bit unsure if we should merge this or disable the scraper for now. The changes itself look fine, but as @balzamas mentioned, the values are "wrong".
I've created https://github.com/openDataBE/covid19Data/issues/1 to see if they can provide the absolute (weekly?) data, as before.

metaodi:
Let's leave this PR for now until we have new information.

PR #1832, opened 2021-07-11
introduce correction_status.csv to skip manual data corrections

maekke:


maekke:
@metaodi @jb3-2 what do you think about this change?

metaodi:
@maekke I'd rather have a clean time series than having another file to ignore some entries. But I don't have strong feelings about this issue, so if you are okay with it, Iet's do it.

maekke:
@metaodi in principle I am agreeing that the time series should be fixed... But NW (and I guess this applies to BS too) states explicit "-2 Fälle aufgrund Systembereinigung".
I just dislike the current situation, that the CSV Validation prints an error every hour...

metaodi:
I totally understand that. In the past I used to remove the obvious wrong values (that were too high and therefore trigger the error) and ignore the values of these days in the scraper (search for "TODO" in the code base and you will find many of those cases). That way at least all the "hacks" are in the same place instead of a new CSV.

jb3-2:
I agree with both of you :-) 

Yes, BS also had some "Systembereinigungen", thus the time series is clean as it is. 

@maekke Thanks for your solution, let's do it!

PR #1841, opened 2021-07-20
JU: scrape Power BI iframe

metaodi:


maekke:
@metaodi I guess this is one approach to do it. I've played around with getting the data via the POST JSON data in https://github.com/openZH/covid_19/pull/1846
It's not yet complete, I haven't really figured out all the responses / types (sometimes the data seems to be computed).

As far as I can tell your scraper should be fine too - feel free to merge this one. Or I can also try to investigate a bit further on the other approach...

metaodi:
I went down the same rabbit hole, and decided to use the selenium approach. Maybe we can convince JU to share the underlying data...

PR #1842, opened 2021-07-20
add logic to detect page with VS district data

### LINKED ISSUE #1837: VS all districts with 0 cases week 27
The numbers are in the pdf, but were not scrapped.

maekke:
Resolves #1837


balzamas:
Thanks a lot!

PR #1846, opened 2021-07-21
WIP: scrape powerbi page for JU data

maekke:


maekke:
@metaodi I've done some additional changes and now scraping seems to basically work... Though I'm not really sure if it's better than the selenium approach (which seems to create some strange values, where the csv validation fails). So I'm fully open to close / ignore this MR for now!
To "complete" this PR, I propose to remove all contents from the csv, otherwise it will not be fully valid.

Issue #1853, opened 2021-07-28
FR Disctricts: week 29 not updated


balzamas:
The cases for week 29 are online (https://www.fr.ch/de/gesundheit/covid-19/coronavirus-statistik-ueber-die-entwicklung-im-kanton) but missing in the csv.

maekke:
Should hopefully be updated again with #1852

balzamas:
This worked, but this week again nothing happened and it seems to me that they once again updated the layout with odd columns ("-7", "-14" etc)
But they have a csv online with daily data, is that new?
https://www.fr.ch/de/document/422471

balzamas:
Broye was imported (8/29), but the other districts failed (last record 8/16).

maekke:
@balzamas with #1886 and #1887 things should (hopefully) be working again. Please let me know, if something is still broken!

balzamas:
I will check next week after the next update, thank you!

balzamas:
It worked, thank you @maekke 

https://twitter.com/CovidlawsStats/status/1442472805656928256?s=20

PR #1886, opened 2021-09-20
fix FR district scraper

maekke:
- use district_xls dict for column matching
- comment html page parsing for now


maekke:
This should resolve #1853

PR #1941, opened 2022-01-21
fix SG hospitalized string

algago:
fix SG hospitalized string from 'Total Patienten mit SARS-CoV-2-Infektion' to 'Total Patienten'


maekke:
@algago thanks a lot for your contribution! :+1:

PR #1963, opened 2022-05-05
pin selenium to version 4.1.3

maekke:
to fix GE and JU scrapers


maekke:
will be fixed with selenium 4.1.5

